PR #14, opened 2023-08-23
Include commits into ZMProof and adjust RO

emmorais:
Include auxiliar commitments into `ZMProof` and adjust verifier to compute the same `x, y, z` as in the prover. 

Added an assert to ensure random oracles are synced.

### DIFF
--- a/.github/workflows/upstream_update.yml
+++ b/.github/workflows/upstream_update.yml
@@ -32,6 +32,31 @@ jobs:
           echo "message=$SCION at $SCION_COMMIT does not include $ROOT at $ROOT_COMMIT" >> $GITHUB_ENV
         fi
       shell: bash
+      
+    - name: Rebase SCION onto ROOT
+      id: rebase
+      run: |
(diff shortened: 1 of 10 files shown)


huitseeker (commented):
@emmorais I'm about to set up CI on this branch (or you can do it yourself by adding it to the triggers), but it would be great to pass `cargo xclippy` on this branch in any case .. by which I mean [not using the dbg macro in checked-in code](https://rust-lang.github.io/rust-clippy/master/index.html#/dbg)

huitseeker (changes requested):
This needs a rebase!

Further, as we've removed the legacy kzg/zeromorph modules, it would be good to scavenge those files : we could use their function definitions to write define tests that support the more efficient `non_hiding_` versions.

huitseeker:
This is integrated in https://github.com/lurk-lab/arecibo/tree/zeromorph as of 6f677be8e1d3e986ea97254554aed27ad13c90f0

PR #16, opened 2023-08-24
clean up secondary circuit

porcuquine:
This PR tries to tease apart usages of the supernova `TrivialTestCircuit`:
- create a new `TrivialSecondaryCircuit` for use in NIVC generally.
- leave the existing test circuit, but remove references to ROM.
- modify uses of program counter on the secondary circuit from the NIVC mechanism, so secondary circuits are not required to provide it
- add some initial support for `CircuitSet` type to formalize what an NIVC user will need to provide

I think it may be possible (and if so, probably desirable) to just use a normal trivial `StepCircuit` (not supernova) for the secondary circuit. However, trying to do this maximally led me into problems I could not debug. Hopefully as the implementation undergoes ongoing crypto review and refactoring/extension it will become clear which parts of the initial scaffolding can be pared back.

### DIFF
--- a/src/supernova/circuit.rs
+++ b/src/supernova/circuit.rs
@@ -76,7 +76,7 @@ pub struct SuperNovaAugmentedCircuitInputs<'a, G: Group> {
   U: Option<&'a [Option<RelaxedR1CSInstance<G>>]>,
   u: Option<&'a R1CSInstance<G>>,
   T: Option<&'a Commitment<G>>,
-  program_counter: G::Base,
+  program_counter: Option<G::Base>,
   last_augmented_circuit_index: G::Base,
 }
 
@@ -91,7 +91,7 @@ impl<'a, G: Group> SuperNovaAugmentedCircuitInputs<'a, G> {
     U: Option<&'a [Option<RelaxedR1CSInstance<G>>]>,
     u: Option<&'a R1CSInstance<G>>,
     T: Option<&'a Commitment<G>>,
-    program_counter: G::Base,
+    program_counter: Option<G::Base>,
     last_augmented_circuit_index: G::Base,
   ) -> Self {
     Self {
@@ -178,7 +178,15 @@ impl<'a, G: Group, SC: StepCircuit<G::Base>> SuperNovaAugmentedCircuit<'a, G, SC
     let program_counter = if self.params.is_primary_circuit {
       Some(AllocatedNum::alloc(
         cs.namespace(|| "program_counter"),
-        || Ok(self.inputs.get()?.program_counter),
+        || {
+          Ok(
+            self
+              .inputs
+              .get()?
+              .program_counter
+              .expect("program_counter missing"),
+          )
+        },
       )?)
     } else {
       None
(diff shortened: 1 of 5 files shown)


src/supernova/circuit.rs
```diff
@@ -178,7 +178,15 @@ impl<'a, G: Group, SC: StepCircuit<G::Base>> SuperNovaAugmentedCircuit<'a, G, SC
...
       Some(AllocatedNum::alloc(
         cs.namespace(|| "program_counter"),
-        || Ok(self.inputs.get()?.program_counter),
+        || {
+          Ok(
+            self
+              .inputs
+              .get()?
+              .program_counter
+              .expect("program_counter missing"),
```
huitseeker, line 187: Why a panic rather than `SynthesisError::AssignementMissing`?

src/supernova/mod.rs
```diff
@@ -659,7 +661,7 @@ where
 
     // secondary circuit
     let num_field_secondary_ro = 2 // params_next, i_new
-    + 2 * pp.F_arity_primary // zo, z1
+    + 2 * pp.F_arity_secondary // zo, z1
```
huitseeker, line 664: Nice 👍

src/supernova/mod.rs
```diff
@@ -810,3 +812,14 @@ where
 
   (ck_primary, ck_secondary)
 }
+
+/// SuperNova helper trait, for implementors that provide sets of sub-circuits to be proved via NIVC.
+pub trait CircuitSet<G: Group> {
```
huitseeker, line 817: Are you looking for an heterogeneous list (or more likely ` ExactSizeIterator`) of `dyn StepCircuit<F>`, where `F: Primefield`?

src/supernova/test.rs
```diff
@@ -178,14 +179,15 @@ where
...
+    pc: Option<&AllocatedNum<F>>,
     z: &[AllocatedNum<F>],
-  ) -> Result<(AllocatedNum<F>, Vec<AllocatedNum<F>>), SynthesisError> {
+  ) -> Result<(Option<AllocatedNum<F>>, Vec<AllocatedNum<F>>), SynthesisError> {
     // constrain rom[pc] equal to `self.circuit_index`
     let circuit_index = alloc_const(
       cs.namespace(|| "circuit_index"),
       F::from(self.circuit_index as u64),
     )?;
+    let pc = pc.unwrap();
```
huitseeker, line 190: An expect wouldn’t go amiss, as it could explain that you intend this `SquareCircuit` to always be used on the primary.

src/supernova/test.rs
```diff
@@ -95,14 +95,15 @@ where
...
```
huitseeker, line 106: An expect would be good, saying you intend the `CubicCircuit` to always be used on the primary.

huitseeker (commented):
This overall LGTM, but I have a few questions inline.

src/supernova/circuit.rs
porcuquine: It's because of the extremely confusing convention of how `Option` is overloaded in the bellman line of circuit-generation (phrased thus because maybe bellpepper will eventually help clean this up). Specifically, if *inputs* are missing (meaning we are synthesizing a blank shape), then the `.get()?` on line 185 will indeed result in the `SynthesisError::AssignmentMissing`. However, if that doesn't happen and inputs have been supplied, then those inputs are either for a primary or secondary circuit. In the secondary case, the program counter will not be present. However, in this code branch, we are specifically handling a primary circuit — so we require the actual value for the witness.

src/supernova/mod.rs
porcuquine: Maybe, but this trait will end up doing some heavy lifting, and it's not really clear yet what it needs to look like. The above is just a first foot in the door. I will iterate in future revisions and bear your ideas in mind.

PR #23, opened 2023-08-26
Refactor to use NonUniformCircuit.

porcuquine:
This PR defines a `NonUniformCircuit` trait which manages most of the complexity of driving NIVC. Applications can provide an implementation of this trait then create proofs as shown in the existing (now refactored) test.

### DIFF
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -12,8 +12,9 @@ use crate::{
   },
   scalar_as_base,
   traits::{
-    circuit_supernova::StepCircuit, commitment::CommitmentTrait, AbsorbInROTrait, Group,
-    ROConstants, ROConstantsCircuit, ROTrait,
+    circuit_supernova::{StepCircuit, TrivialSecondaryCircuit},
+    commitment::CommitmentTrait,
+    AbsorbInROTrait, Group, ROConstants, ROConstantsCircuit, ROTrait,
   },
   Commitment, CommitmentKey,
 };
@@ -28,6 +29,7 @@ use crate::bellpepper::{
 };
 use bellpepper_core::ConstraintSystem;
 
+use crate::compute_digest;
 use crate::nifs::NIFS;
 
 mod circuit; // declare the module first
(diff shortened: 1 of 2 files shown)


src/supernova/mod.rs
```diff
@@ -813,13 +815,74 @@ where
...
+
+    let running_claim_params = running_claims
+      .iter()
+      .map(|c| c.get_public_params())
+      .collect::<Vec<_>>();
+
+    let (ck_primary, ck_secondary) = compute_commitment_keys(&running_claim_params);
+
+    for claim in &mut running_claims {
+      claim.set_commitment_key(ck_primary.clone(), ck_secondary.clone());
```
huitseeker, line 865: It seems quite essential to make sure that the `params` field of `RunningClaim` is an `Rc` to a `PublicParams` singleton, this sort of duplication is quickly going to be unmanageable.

Note the only reason I could see that we're manipulating a `Vec<RunningClaim<...>>` is to be able to index into them. We could have a `struct RunningClaim<...>` which would host the public parameter reference and some `impl Index`, `impl IndexMut` impls. Shall we open an issue?

src/supernova/mod.rs
```diff
@@ -813,13 +815,74 @@ where
...
-  fn initial_program_counter(&self) -> G::Scalar {
-    G::Scalar::ZERO
+  fn initial_program_counter(&self) -> G1::Scalar {
+    G1::Scalar::ZERO
+  }
+
+  /// Return the initial running claims for `NonUniformCircuit`'s sub-circuits.
+  fn initial_running_claims(
+    &self,
+  ) -> Vec<RunningClaim<G1, G2, C1, TrivialSecondaryCircuit<G2::Scalar>>> {
```
huitseeker, line 835: By 1.74, we should be able to make this return an `impl IntoIterator<Item = ...>`, instead of this form, which forces an allocation of an intermediate result that's never used on its own. The feature is `return_position_impl_trait_in_trait`

src/supernova/test.rs
```diff
@@ -276,26 +273,74 @@ fn print_constraints_name_on_error_index<G1, G2, Ca, Cb>(
...
+{
+  fn num_circuits(&self) -> usize {
     2
   }
+
+  fn primary_circuit(&self, circuit_index: usize) -> TestRomCircuit<G1::Scalar> {
+    match circuit_index {
+      0 => TestRomCircuit::Cubic(CubicCircuit::new(circuit_index, self.rom.len())),
+      1 => TestRomCircuit::Square(SquareCircuit::new(circuit_index, self.rom.len())),
+      _ => unimplemented!(),
```
huitseeker, line 327: Nit: Is `unimplemented` the right approach, considering `num_circuits` states plainly the call is invalid?

huitseeker (approved):
Nothing blocking, but I left a few comments inline.

src/supernova/mod.rs
porcuquine: Yes, agreed — please create an issue.

huitseeker: https://github.com/lurk-lab/arecibo/issues/29

huitseeker (approved):

PR #32, opened 2023-08-30
chore: Clean up Microsoft licensing

samuelburnham:
Removes confusion relating to Microsoft licensing/policies. I'm not exactly sure what the right protocol is for a long-lived fork, please suggest any changes.


huitseeker (changes requested):
We actually need to reproduce the MS license, it's against the license terms to remove it. 
The correct way is to indeed insert our (c) but move Microsoft's to a 3d-party notice file.
See https://github.com/lurk-lab/arecibo/commit/d58bf7ed5220a031238026002390fb75f0b4b198 for an example (on a different branch of this repo)

ThirdPartyNotices.txt
```diff
@@ -0,0 +1,54 @@
...
+------------------------------------------------------------
+https://github.com/espressosystems/jellyfish
+https://github.com/han0110/plonkish/
+https://github.com/espressosystems/hyperplonk
+
+Licensed under MIT
+
+MIT License
+
+Copyright (c) 2022 Espresso Systems, han0110
```
huitseeker, line 41: This branch technically contains none of these but eh, the discrepancy should self correct soon enough.

huitseeker (approved):

PR #34, opened 2023-08-31
Add a `digest` field to `R1CSShape`

### LINKED ISSUE #31: Add a digest to circuits
From: https://zulip.lurk-lab.com/#narrow/stream/22-inc/topic/Push.20to.20Coprocessors/near/109810

porcuquine: 
> Unfortunately, once the constraints change, the public params for the coproc will fall out of sync. We don't do any smarter checking as of now. 

Now that we are getting into coprocessors, it's fairly essential that we fix this. It's going to eat up an increasing amount of the team's time the longer it goes unaddressed. I suggest we include the digest from the Nova params in the cache key — or something morally equivalent. I haven't thought about this since previously, but in poking at SuperNova, I've seen that the digest does now exist, so…

porcuquine: The most important point is that we need a digest of the circuit(s) so that if a circuit changes in any way the public params will also change.

porcuquine: The easiest way to do this probably involves using Nova's ability to compute a digest over the pp components, including the circuit shapes. It might be that we can/should add some trait that lets us get a 'partial digest' sufficient for our purposes to arecibo, then use that.

winston-h-zhang:
Resolves #31 and https://github.com/lurk-lab/lurk-rs/pull/648 downstream.

This PR adds a small mechanism to distinguish circuits via a digest. This is essentially the same as what is done for `PublicParams.digest`. Downstream in `lurk-rs`, the digest is needed to quickly determine if a set of public params matches the given circuits. Note that the cost of `circuit_digest` is somewhat small compared to fully generating params, but it remains to be benchmarked. If the cost is too high, maybe something more clever will be needed.

### DIFF
--- a/src/digest.rs
+++ b/src/digest.rs
@@ -22,24 +22,34 @@ pub trait HasDigest<F: PrimeField> {
 
 /// Trait for components with potentially discrete digests to be included in their container's digest.
 pub trait Digestible {
-  /// Write the byte representation of Self in a byte buffer
-  fn write_bytes<W: Sized + io::Write>(&self, byte_sink: &mut W) -> Result<(), io::Error>;
+  /// Write the digest representation of Self in a byte buffer
+  /// this is what you want to write when Self is being digested into another digest
+  fn write_digestable_bytes<W: Sized + io::Write>(
+    &self,
+    byte_sink: &mut W,
+  ) -> Result<(), io::Error>;
   /// allocate and exhibit the bytes for the type in question
-  fn to_bytes(&self) -> Result<Vec<u8>, io::Error> {
-    let mut v: Vec<u8> = Vec::new();
-    self.write_bytes(&mut v)?;
-    Ok(v)
-  }
+  /// this is what you want when you need the representation of Self to compute its digest
+  fn to_bytes(&self) -> Result<Vec<u8>, io::Error>;
 }
 
 /// Marker trait to be implemented for types that implement `Digestible` and `Serialize`.
 /// Their instances will be serialized to bytes then digested.
 pub trait SimpleDigestible: Serialize {}
 
+/// In the case where the digest is "simple", the digest is the byte representation
 impl<T: SimpleDigestible> Digestible for T {
-  fn write_bytes<W: Sized + io::Write>(&self, byte_sink: &mut W) -> Result<(), io::Error> {
+  fn write_digestable_bytes<W: Sized + io::Write>(
+    &self,
+    byte_sink: &mut W,
+  ) -> Result<(), io::Error> {
     bincode::serialize_into(byte_sink, self)
       .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))
+  }
+  fn to_bytes(&self) -> Result<Vec<u8>, io::Error> {
+    let mut v: Vec<u8> = Vec::new();
+    self.write_digestable_bytes(&mut v)?;
+    Ok(v)
   }
 }
 
@@ -82,10 +92,16 @@ impl<F: PrimeField, T: HasDigest<F> + Digestible> DigestBuilder<F, T> {
   /// Build and return inner `Digestible`.
   pub fn build(mut self) -> Result<T, io::Error> {
     let mut hasher = Self::hasher();
-    self.inner.write_bytes(&mut hasher)?;
+    let bytes = self.inner.to_bytes()?;
+    hasher.update(&bytes);
     let mut bytes: [u8; 32] = hasher.finalize().into();
+
     self.inner.set_digest(Self::map_to_field(&mut bytes));
 
     Ok(self.inner)
   }
 }
+
+impl SimpleDigestible for usize {}
+impl<T> SimpleDigestible for PhantomData<T> {}
+impl<T: Serialize> SimpleDigestible for Vec<T> {}
--- a/src/digest.rs
+++ b/src/digest.rs
@@ -1,3 +1,4 @@
+//! This module provides a unified interface for working with digests of this libraries data.
 use digest::{typenum::Unsigned, OutputSizeUser};
 use ff::PrimeField;
 use serde::Serialize;
@@ -70,7 +71,8 @@ impl<F: PrimeField, T: HasDigest<F> + Digestible> DigestBuilder<F, T> {
     Sha3_256::new()
   }
 
-  fn map_to_field(digest: &mut [u8]) -> F {
+  /// Hash an arbitrary amount of bytes to a field element
+  pub fn map_to_field(digest: &mut [u8]) -> F {
     let bv = (0..NUM_HASH_BITS).map(|i| {
       let (byte_pos, bit_pos) = (i / 8, i % 8);
       let bit = (digest[byte_pos] >> bit_pos) & 1;
@@ -102,6 +104,7 @@ impl<F: PrimeField, T: HasDigest<F> + Digestible> DigestBuilder<F, T> {
   }
 }
 
+impl SimpleDigestible for bool {}
 impl SimpleDigestible for usize {}
 impl<T> SimpleDigestible for PhantomData<T> {}
 impl<T: Serialize> SimpleDigestible for Vec<T> {}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1002,6 +1002,64 @@ mod tests {
     }
   }
 
+  fn test_circuit_digest_with<G1, G2, T1>(circuit: &T1, expected: &str)
+  where
+    G1: Group<Base = <G2 as Group>::Scalar>,
+    G2: Group<Base = <G1 as Group>::Scalar>,
+    T1: StepCircuit<G1::Scalar>,
+  {
+    let digest = circuit_digest::<G1, G2, _>(circuit, true);
+
+    let digest_str = digest
+      .to_repr()
+      .as_ref()
+      .iter()
+      .map(|b| format!("{b:02x}"))
+      .collect::<String>();
+    assert_eq!(digest_str, expected);
+  }
+
+  #[test]
+  fn test_circuit_digest() {
+    type G1 = pasta_curves::pallas::Point;
+    type G2 = pasta_curves::vesta::Point;
+    let cubic_circuit1 = CubicCircuit::<<G1 as Group>::Scalar>::default();
+    let cubic_circuit2 = CubicCircuit::<<G2 as Group>::Scalar>::default();
+
+    test_circuit_digest_with::<G1, G2, _>(
+      &cubic_circuit1,
+      "789427954a51e446fa7905c64242efba6405d06d2d5f8d044fab5473e04ca802",
+    );
+
+    test_circuit_digest_with::<G2, G1, _>(
+      &cubic_circuit2,
+      "4962a68550cd0a387c594152684d71008648f51a7407a1708a31becd797d9303",
+    );
+
+    let cubic_circuit1_grumpkin = CubicCircuit::<<bn256::Point as Group>::Scalar>::default();
+    let cubic_circuit2_grumpkin = CubicCircuit::<<grumpkin::Point as Group>::Scalar>::default();
+
+    test_circuit_digest_with::<bn256::Point, grumpkin::Point, _>(
+      &cubic_circuit1_grumpkin,
+      "158c106e5cd7156050fbd585c4d624daa92288b72b979ca296ac1865dc8eae03",
+    );
+    test_circuit_digest_with::<grumpkin::Point, bn256::Point, _>(
+      &cubic_circuit2_grumpkin,
+      "8d30e95b9decf300b9eaaf9304088bce5a04af3d952cedc81a7735fc049f8203",
+    );
+
+    let cubic_circuit1_secp = CubicCircuit::<<secp256k1::Point as Group>::Scalar>::default();
+    let cubic_circuit2_secp = CubicCircuit::<<secq256k1::Point as Group>::Scalar>::default();
+    test_circuit_digest_with::<secp256k1::Point, secq256k1::Point, _>(
+      &cubic_circuit1_secp,
+      "0442eeb939c179f7d7cc83af96af742ab1a3bcf85d2633f2a0ef2547105a5602",
+    );
+    test_circuit_digest_with::<secq256k1::Point, secp256k1::Point, _>(
+      &cubic_circuit2_secp,
+      "f4911e3149c51ed48440dc93996cdb111c33750d319a6f297e740cef1afa2700",
+    );
+  }
+
   fn test_pp_digest_with<G1, G2, T1, T2>(circuit1: &T1, circuit2: &T2, expected: &str)
   where
     G1: Group<Base = <G2 as Group>::Scalar>,
@@ -1039,13 +1097,13 @@ mod tests {
     test_pp_digest_with::<G1, G2, _, _>(
       &trivial_circuit1,
       &trivial_circuit2,
-      "39a4ea9dd384346fdeb6b5857c7be56fa035153b616d55311f3191dfbceea603",
+      "cc9b8776b7b04ba04daa6c5dc064384686a700921d4411d2a806261b41d01102",
     );
 
     test_pp_digest_with::<G1, G2, _, _>(
       &cubic_circuit1,
       &trivial_circuit2,
-      "3f7b25f589f2da5ab26254beba98faa54f6442ebf5fa5860caf7b08b576cab00",
+      "0cc25f7bedbb01a3d092f004f721c0536754cfffda4d5c125d3aaffec3af9b03",
     );
 
     let trivial_circuit1_grumpkin =
@@ -1057,12 +1115,12 @@ mod tests {
     test_pp_digest_with::<bn256::Point, grumpkin::Point, _, _>(
       &trivial_circuit1_grumpkin,
       &trivial_circuit2_grumpkin,
-      "967acca1d6b4731cd65d4072c12bbaca9648f24d7bcc2877aee720e4265d4302",
+      "1c1f7e860dcb0ce9501cd86dbfbfcaa4aa6cfa1ff55421e9f5e2c668224f8c03",
     );
     test_pp_digest_with::<bn256::Point, grumpkin::Point, _, _>(
       &cubic_circuit1_grumpkin,
       &trivial_circuit2_grumpkin,
-      "44629f26a78bf6c4e3077f940232050d1793d304fdba5e221d0cf66f76a37903",
+      "b526eae3c3c5d353f99d1b4e097048b3114dcbdbdceb030cbae20de7aafd8b01",
     );
 
     let trivial_circuit1_secp =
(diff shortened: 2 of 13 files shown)


src/r1cs.rs
```diff
@@ -347,14 +355,19 @@ impl<G: Group> R1CSShape<G> {
...
         num_cons: m,
         num_vars: m,
         num_io: self.num_io,
         A: self.A.clone(),
         B: self.B.clone(),
         C: self.C.clone(),
+        digest: G::Scalar::ZERO,
       };
+
+      // it's a shame we have to recompute everything here; perhaps we can optimize this somehow
```
huitseeker, line 368: We will indeed need that optimization, because the same bytes will be hashed again in https://github.com/lurk-lab/arecibo/blob/c3608280054fd183c65873b88251e2710d483c3f/src/lib.rs#L203-L204

Evidence that this is performance-sensitive exists in https://github.com/microsoft/Nova/pull/182/files, which introduced the benchmark and tests for the public parameter's digest.

huitseeker (dismissed):
From what I understand of https://github.com/lurk-lab/arecibo/issues/31 and the downstream https://github.com/lurk-lab/lurk-rs/pull/648, you need a hash of the public parameters that changes every time the circuits involved in those public parameters change, and can be used as a public parameter cache key so that no one can load the wrong public parameters for the current circuits?

If I have understood that requirement correctly, then good news! that digest already exists, it's [the public parameter digest](https://github.com/lurk-lab/arecibo/blob/c3608280054fd183c65873b88251e2710d483c3f/src/lib.rs#L868-L895), it was introduced in https://github.com/microsoft/Nova/pull/168 and it is cached in the public parameters already. No need to add anything : if the circuits change, the public parameter digest will change correspondingly.[^1]

Now, it's possible I missed something, just lmk.


[^1]: Now you'll tell me that the public parameters that are used for Supernova don't include the full range of circuits used in the accumulator**s**: they contain but the largest instance, and not the full set. I would argue that this is *plain wrong*, in the sense that it invalidates the purpose of https://github.com/microsoft/Nova/pull/168, which is to bind the prover to a specific set of accumulators.
In that sense, *even if the public parameters only need the shape of the largest circuit for most of the time*, they should at least pass *the whole ROM* (all accumulators) at the time of populating the digest. The pattern to use there, I think, is a builder:
https://rust-unofficial.github.io/patterns/patterns/creational/builder.html -and there the builder would receive the ROM and possibly not keep it in the PPs post vomiting the hash. Or (and I'd probably prefer this) keep the whole array of R1CS shapes in the PPs.
Fortunately, we can tackle all this a bit later, because, as you probably noticed, there are two sets of public parameters, [one for Nova](https://github.com/lurk-lab/arecibo/blob/c3608280054fd183c65873b88251e2710d483c3f/src/lib.rs#L69-L92) (which is fine), [one for Supernova](https://github.com/lurk-lab/arecibo/blob/c3608280054fd183c65873b88251e2710d483c3f/src/supernova/mod.rs#L51-L68) (which I have strong issues with, as laid out in the above).

/cc @porcuquine

porcuquine:
> If I have understood that requirement correctly, then good news! that digest already exists, it's [the public parameter digest](https://github.com/lurk-lab/arecibo/blob/c3608280054fd183c65873b88251e2710d483c3f/src/lib.rs#L868-L895), it was introduced in [microsoft/Nova#168](https://github.com/microsoft/Nova/pull/168) and it is cached in the public parameters already. No need to add anything : if the circuits change, the public parameter digest will change correspondingly.

I believe the issue is that we need not only a digest that changes whenever the parameters do but one that can also be computed *without* incurring the major cost of actually generating the parameters. Specifically, we need this as a cache key exactly in order to avoid that.

We can think of there being several parts:
- Very expensive parts: commitment keys, primarily, if I'm not mistaken
- Application-specific 'differentiating' parts: critically, the `R1CSShape`s.
- Somewhat expensive parts: the random oracle constants,  which must be (inefficiently, for now) generated by neptune
- Cheap parts: the rest, more or less
- Nova-specific parameters: technically, the arity of the random oracle hash could be different than what is currently deterministically computed, as could the limbs, etc.

Since we are not implementing the parameter caching as an arecibo feature, we should primarily consider a solution that addresses the 'application-specific' needs. Ideally this can be done in a way that is not intrusive and incurs no performance penalty for applications that don't make use of this.

In other words, I think it is indeed the case that we need to export the ability to get a digest just from the application-specific part: the `R1CSShape`; while avoiding the expensive parts (the point of the cache): the commitment keys and random oracle constants.

For NIVC, this will need to include *all* R1CS shapes, as you note. For downstream purposes, it may be acceptable to compose such a composite digest from a canonical digest of the individual `R1CShape`s, though. As likely, we may choose a solution that allows the shape part of the overall digest to be shared between 'consumers' and the folding scheme itself.

porcuquine:
> 1. Now you'll tell me that the public parameters that are used for Supernova don't include the full range of circuits used in the accumulator**s**: they contain but the largest instance, and not the full set. I would argue that this is _plain wrong_, in the sense that it invalidates the purpose of https://github.com/microsoft/Nova/pull/168, which is to bind the prover to a specific set of accumulators.
>    In that sense, _even if the public parameters only need the shape of the largest circuit for most of the time_, they should at least pass _the whole ROM_ (all accumulators) at the time of populating the digest.

To make sure we're on the same page, let me note that the ROM is a concept only used in the toy 'NIVC application' used in the SuperNova tests as part of the initial draft. It's essentially a static representation of the NIVC 'program' to be proved. I don't think this distinction is significant to the discussion, but I mention it so we can clarify any confusion about terminology.

I agree that all shapes must be accounted for in the parameter digest for SuperNova. As noted above, I believe that is already the case. I also agree that neither the creation method nor precise content of those parameters is optimal. Finally, I suspect that once we have reached a better version of creation/content there will also be a better method of creating the joint digest than the current.

src/r1cs.rs
arthurpaulino: Can we compute it only if it's equal to `G1::Scalar::ZERO` (the default initial value)?

winston-h-zhang: Padding occurs as a one time cost at the end when producing the snark. Technically we currently don't even need to check any digest at this point, since it's assumed that the public params are already loaded and valid (though in the future it might be necessary to check again due to some new complexity). So my inclination to to delete this line and add a comment making it clear that this is "wrong," but we don't care.

Indeed, this is pretty compute heavy so we have to be careful.

winston-h-zhang: @arthurpaulino we could compute it for any default initial value (like `37`, or `None` with `Option<F>`). We just had to pick a default one to be able to serialize the structure (and make sure it's the same default each time).

porcuquine:
I think this can now be reworked, to use `HasDigest`, `Digestible`, `DigestBuilder` from #40. Because of https://github.com/lurk-lab/arecibo/pull/40#issuecomment-1711986370, this can't be as clean as would be ideal — but we can start with manual implementation of `Digestible::write_bytes()` allowing the circuit to be digested only once, with the digest itself used when creating the digest for the containing parameters.

winston-h-zhang:
@porcuquine Rebasing on #40, I've hacked around the `SimpleDigestable` conflict by separating the use of `write_bytes` (or now `write_digest`) to be the representation of what should be digested within the context of a larger container, and `to_bytes` to be the representation of what should be computed to generate the digest. For `SimpleDigestable`, the two are the same. As you mentioned I manually implemented `Digestable` for the public params and running claims.

src/lib.rs
```diff
@@ -208,13 +208,45 @@ where
...
   G2: Group<Base = <G1 as Group>::Scalar>,
   C1: StepCircuit<G1::Scalar>,
   C2: StepCircuit<G2::Scalar>,
+  G1::Scalar: Serialize,
 {
+  fn write_digest<W: Sized + io::Write>(&self, byte_sink: &mut W) -> Result<(), io::Error> {
+    self.digest.write_digest(byte_sink)
+  }
+
+  fn to_bytes(&self) -> Result<Vec<u8>, io::Error> {
```
winston-h-zhang, line 223: Perhaps we could implement a quick proc macro for this

src/digest.rs
```diff
@@ -22,25 +22,29 @@ pub trait HasDigest<F: PrimeField> {
 
 /// Trait for components with potentially discrete digests to be included in their container's digest.
 pub trait Digestible {
-  /// Write the byte representation of Self in a byte buffer
-  fn write_bytes<W: Sized + io::Write>(&self, byte_sink: &mut W) -> Result<(), io::Error>;
+  /// Write the digest representation of Self in a byte buffer
+  /// this is what you want to write when Self is being digested into another digest
+  fn write_digest<W: Sized + io::Write>(&self, byte_sink: &mut W) -> Result<(), io::Error>;
```
porcuquine, line 27: I think this may need a better name. I think the separated functions make sense, but for example `SimpleDigestible` implements this function in a way that does not involve writing it's *digest*.

Maybe this can be called `write_summary_bytes` or something? That is not a well-thought-out suggestion, but the point is that this function might either write the digest or the input over which the digest was created — depending on context. At least that is what I've understood.

src/r1cs.rs
```diff
@@ -354,14 +385,22 @@ impl<G: Group> R1CSShape<G> {
...
         num_io: self.num_io,
         A: self.A.clone(),
         B: self.B.clone(),
         C: self.C.clone(),
+        // TODO: at this point within the proving pipeline, we assume that `R1CSShape::pad`
+        // is called only in the ppsnark/snark setup process. Thus, the digest's use as a check
+        // to prevent drifting public params is not a concern, and we can avoid doing a massive
+        // recomputation for a new digest that will never be used. However, this is ad-hoc and
+        // we should find a nicer solution
+        digest: self.digest,
```
winston-h-zhang, line 400: Another concern. As I wrote this, I realized that currently `master` doesn't take into this concern either, but it is more subtle. After padding, the digest of the public params will not match the new padded shape. Looking at the setup code in `snark.rs`/`ppsnark.rs`, I see that we commit again to the padded shape though, so this shouldn't be an issue, right? 

cc: @huitseeker

src/digest.rs
winston-h-zhang: That does make sense; maybe `write_digestable_bytes` to indicate that these are the bytes being digested (into another container)?

porcuquine: Yes, I think that's the ticket.

huitseeker (changes requested):
I think I understand what you're trying to do, that is you want:
- the `Digestible` trait's main functionality should be to stream bytes into an `io::Write`r passed as argument,
- an implementation of the `Digestible` trait by which, for objects that are serializable, you stream the bytes of their serialized form. That's exactly the `SimpleDigestible` implementation, or at least the `write_digestable_bytes` (sic) of it. To recall, the functionality of this trait should be to inject a stream of bytes in a `Writer`
- an extension trait of `Digestible`, say `Hashable` by which you can equip the `Digestible` objects with a `digest()` function (or equivalent). The functionality of this trait should by to compute an actual hash from the `Digestible` bytes.
- a proc-macro by which you can *derive* the `Digestible` trait on an object O, so that for each field of that object, assuming those fields implement the `Hashable` trait, you can stream:
- a marker of the type of the object, which should be a domain separator (the [`serde_name`](https://docs.rs/serde-name/latest/serde_name/) crate would help),
- the *hashed* bytes of each field, 
- *except for digest field of O*. 
To recall, the functionality of this trait (`Digestible`) should also be to inject bytes in a `Writer`.

One goal of this would be not to allow any hash collisions between types that stream their ground serialization bytes, and those that stream their hash bytes.

src/r1cs.rs
huitseeker: The point of the builder pattern I produced in #31
Was to make sure you'd run all the initialization steps before getting the object, and therefore did not have to check an object to see if it had been padded or not. Did we forget padding as part of this ?

winston-h-zhang: I don't understand what you mean. My point is that after the digest is built, we mutate the shape by padding it, and we don't rebuild the new padded shape to update the digest. However, because the new digest wouldn't be used anyway, since the transcript implicitly consumes the digest of the unpadded shape through pp.digest. Hence we can elide the expensive recompute.

However, I don't know if this may raise ways to attack the snark protocol.

src/lib.rs
winston-h-zhang: `to_bytes` is what creates a digest from a raw object `T`. `write_digestible_bytes` is how `T` should be represented when creating the digest of a larger object `S` can contains `T`. For `R1CSShape` and `PublicParams` we break the invariant because when computing the digest of `PublicParams` that contains `R1CSShape`s, we simply want to absorb `shape.digest` instead of rehashing all of `shape` again. Thus the implementation of `write_digestible_bytes` for `R1CSShape` should just write its digest and nothing else.

src/lib.rs
```diff
@@ -208,13 +208,60 @@ where
...
+    self
+      .r1cs_shape_secondary
+      .write_digestable_bytes(&mut byte_sink)?;
+    self
+      .augmented_circuit_params_primary
+      .write_digestable_bytes(&mut byte_sink)?;
+    self
+      .augmented_circuit_params_secondary
+      .write_digestable_bytes(&mut byte_sink)?;
+    self.digest.write_digestable_bytes(&mut byte_sink)?;
```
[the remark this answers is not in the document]
winston-h-zhang: :O nice catch

winston-h-zhang:
I'm a bit confused by your suggestions for `Hashable` to further extend `Digestible`. I feel like we essentially need two APIs for any `T`:

1. Compute `T -> Digest`
2. Given `T < S`, how to include `T` into `S`. If `T.digest` exists, then we include `T.digest`, otherwise we include the bytes of `T`. Or in general some arbitrary function `digestible_repr(T)` that the user can implement.

Given this, we could have 2 traits for each point, instead of 1 trait with two functions (plus `HasDigest` and possibly `Hashable`). @huitseeker let me know what you think!

src/lib.rs
```diff
@@ -208,13 +208,60 @@ where
...
+    self._p_c1.write_digestable_bytes(&mut byte_sink)?;
+    self._p_c2.write_digestable_bytes(&mut byte_sink)?;
```
huitseeker, line 262: Another important point: do you think this actually binds the public parameters to the shape of the circuit?
Because phantom types beg to differ:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ffd3e4adc7538effc0a088b1b0474734

huitseeker (commented):
@winston-h-zhang The digest computed here is security-critical, so we need to be more precise in specifying it.  There's no such thing as a function that can be specified with as few words as `T -> Digest` and I don't know how to parse what `T < S` should mean.

In order to define a digest, you need:
1. the actual hash computation, which involves a hasher that can munge bytes. It should not require allocating these bytes in e.g. a vector.,
2. the selection of the bytes to be added as argument to (1.),

Things we should do:
- define statically an unambiguous way of getting the bytes that serve as an input for a hasher, for a large class of objects,
- *not* define any other way of getting any other bytes that could possibly be construed to be a suitable input for that very hasher,

With that said, I understand the purpose of this PR is to also offer an indirection by selecting the digests of constituent fields as inputs, so as to avoid re-computing the hash of those embedded objects. 

The reason I suggest splitting this in two traits is that part (1.) above is the same for all objects, irrespective of their constituents, and (2.) is not.

But this goes a bit further: So as to make the meaning of what we're hashing precise, we also need to use domain separation when inserting those hashes, so that there is no possible way of creating a hash collision between the following objects:
```rust
mod foo {
  struct S ( [u8; 32] );
  impl SimpleDigest for [u8; 32] {}
}
// and
mod bar {
  struct S ( [u8; 32] );
  impl SimpleDigest for S {}
}
```
This will require at least using `serde_name` as indicated above, and adding a domain separator to the addition of the digest in the hasher's input, something I believe your approach to `Digestible` does not do today (and I have trouble imagining how to make it do so).

For a glimpse of why this is important, see https://crypto.junod.info/posts/recursive-hash/

Please schedule some time with me so we can go over this in depth.

Notes: [1] [2]

[1]: Since our hashers, including sha3, are quite fast, it's actually quite fine to start an implementation by just *not using digests as inputs* to the hasher (in 2. above), thereby flattening the bytes involved in the digest computation to the serialization bytes of the structure. We have two desires:
a. speeding up the equality comparison in the case of embedded structures, e.g. when we have two `PublicParameters` that each embed circuits that are known to be distinct,
b. speeding up the computation of overall hash of the top-level structure, by reusing the hash of the imbricated structure, rather than re-computing it.
Note that a. comes quite ahead of b. in priority. (b.) is a secondary concern and a mere performance optimization. I believe the following is an acceptable option to start iterating from, if sub-optimal from a perf standpoint:
```rust
fn cache_key(pp: PublicParams<...>) -> String {
  let str1 = pp.r1cs_shape_primary.to_bytes(); // full serialization and hashing of r1cs
  let str2 = pp.to_bytes(); // full serialization and hashing of pp, including the second serialization & hashing of r1cs
  format!("{str1}-{str2}") 
}
```
Then we can write the interesting part (a. above): a function that reads the pp from cached files only if the r1cs prefix matches.


[2]: Prior to this PR, I wrote the Digestible trait with two functions: `write_bytes`, and the decorative, documentary `to_bytes`. I now realize with this PR that adding the second function was a mistake, because it is giving you ideas about giving a funny meaning to `to_bytes`.  I think we should remove it. I'm also in favor of changes that involve redefining the trait hierarchy ... as long as it makes rigorous sense.

src/lib.rs
porcuquine: I think these can be left out, exactly because they are *not* actually 'present'. The shapes are there independently, if I'm not mistaken — as straightforward data: https://github.com/lurk-lab/arecibo/blob/b4f3ce80980e6a9f985b5fe9ec44f37b1e21d267/src/lib.rs#L198-L202

porcuquine:
> With that said, I understand the purpose of this PR is to also offer an indirection by selecting the digests of constituent fields as inputs, so as to avoid re-computing the hash of those embedded objects.

I hesitate to interject, because I have not tracked this discussion entirely. However, I want to make sure I'm following. I understand generally what @winston-h-zhang is going for. I *think* what @huitseeker describe above is slightly different. At least the wording of this sentence suggests that.

The nuance I observe is that in @winston-h-zhang's plan, there are two distinct ways of ensuring that a constituent object is included in a containing digest.

- In the simple case, the raw bytes of that object are hashed into the hasher state of the container. In this case, the 'digest' or specific hash of those bytes, is never actually computed independently.
- Alternately, *when the contained object itself has a digest* (ignoring for now the specific trait that will trigger this — as I would have to refresh my memory of the discussion) — then instead the bytes of the *digest itself* will be hashed into the container's digest.

In this model, the contained object's digest will always be computed exactly once. However, inclusion of that digest as part of the input to be digested in the container is not directly equivalent to the first instance. It still binds the parent's hash to the content of the contained object — but it will not result in the same final digest. I believe this is fine.

It might be that this was absolutely clear to both parties already. In that case, carry on. If not, maybe this helps in some way.

huitseeker:
@porcuquine thanks for clarifying the goals, I believe that we are all on the same page as to what they are. They are also clear in the code. I'm just highlighting that performing the second case (hashing digests recursively) safely (without opening ourselves to collision attacks) requires rigor and care. 

And my note [1] suggests that until we have defined a rigorous and safe way to perform this in code (something that will indeed have advantages from a performance standpoint), perhaps we could limit ourselves to the somewhat more modest goal of computing digests strictly from serialized forms (not digests themselves), and simply using the digest of a constituent field as a prefix of a cache key when this field's digest is expected to  be enough to discriminate between two sets of public parameters

src/lib.rs
huitseeker: Yes, that's what I was getting at.

src/lib.rs
huitseeker: But now for all structures that use the digests of their constituent fields (rather than the serialization bytes of their constituent fields), I have two ways of getting to a digest for the top-level structure: hashing the output of `to_bytes`, and asking `write_digestable_bytes` to write the digest to a byte sink. 

For some structures this is the same. For some this is not. This ambiguity is bound to end in tears.

winston-h-zhang:
Closed in favor of #49

PR #38, opened 2023-09-05
CSR Representation for Sparse Matrices

### LINKED ISSUE #37: Replace COO SpMVM by a CSR
https://github.com/lurk-lab/arecibo/pull/36 clearly shows that an SpMVM (Sparse Matrix-Vector Multiplication) is responsible for a significant amount of the folding time:
https://github.com/lurk-lab/arecibo/blob/c3608280054fd183c65873b88251e2710d483c3f/src/r1cs.rs#L169-L172

This operation is done on matrices in COO form, whereas the typical representation for such operations would be a CSR form ([wiki](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format))). These matrices are created in the synthesis of `(Test)ShapeCS` :
https://github.com/lurk-lab/arecibo/blob/c3608280054fd183c65873b88251e2710d483c3f/src/bellpepper/r1cs.rs#L103-L104

Since those matrices are used only for SpMVM, we hope to replace all instances of COO by a CSR (unless something unexpected turns up, such as patterns of random access in the A, B, C matrices). 

The reason that we care about this representation is that it will make all the matrix elements relevant in a row of matrix-vector multiplication adjacent in memory, which opens a path to more performant code, amenable to LLVM's auto-vectorizer.

winston-h-zhang:
Resolves #37.

This PR implements CSR matrix-vector multiplication under the `"csr"` feature. (This was for me to debug; if CSR sould become the only representation, then we can simply replace and remove the feature.) We get a slight uptick in performance, around 10%.

Notes:

There are many sparse matrix crates available in rust. However, most of these crates are do not parallelize, or are not compatiable with field elements. For example, [sprs](https://crates.io/crates/sprs/0.2.1) does not parallelize its sparse matrix / dense vector multiplication. It's traits also require `num_traits::Zero`, which is not compatiable with `ff::PrimeField`.

CSR:
```
Nova-based VDF with MinRoot delay function
=========================================================
Proving 1 iterations of MinRoot per step
Producing public parameters...
PublicParams::setup, took 898.236583ms 
Number of constraints per step (primary circuit): 9819
Number of constraints per step (secondary circuit): 10347
Number of variables per step (primary circuit): 9813
Number of variables per step (secondary circuit): 10329
Generating a RecursiveSNARK...
RecursiveSNARK::prove_step 0: true, took 3.208µs 
RecursiveSNARK::prove_step 1: true, took 31.0505ms 
RecursiveSNARK::prove_step 2: true, took 37.113ms 
RecursiveSNARK::prove_step 3: true, took 37.011ms 
RecursiveSNARK::prove_step 4: true, took 38.116292ms 
RecursiveSNARK::prove_step 5: true, took 37.588333ms 
RecursiveSNARK::prove_step 6: true, took 37.468125ms 
RecursiveSNARK::prove_step 7: true, took 37.913125ms 
RecursiveSNARK::prove_step 8: true, took 37.48625ms 
RecursiveSNARK::prove_step 9: true, took 38.424084ms 
Verifying a RecursiveSNARK...
RecursiveSNARK::verify: true, took 29.615583ms
Generating a CompressedSNARK using Spartan with IPA-PC...
CompressedSNARK::prove: true, took 721.547ms
CompressedSNARK::len 9407 bytes
Verifying a CompressedSNARK...
CompressedSNARK::verify: true, took 30.767459ms
```

No CSR:
```
Nova-based VDF with MinRoot delay function
=========================================================
Proving 1 iterations of MinRoot per step
Producing public parameters...
PublicParams::setup, took 891.275666ms 
Number of constraints per step (primary circuit): 9819
Number of constraints per step (secondary circuit): 10347
Number of variables per step (primary circuit): 9813
Number of variables per step (secondary circuit): 10329
Generating a RecursiveSNARK...
RecursiveSNARK::prove_step 0: true, took 2.875µs 
RecursiveSNARK::prove_step 1: true, took 34.667833ms 
RecursiveSNARK::prove_step 2: true, took 40.504625ms 
RecursiveSNARK::prove_step 3: true, took 40.93575ms 
RecursiveSNARK::prove_step 4: true, took 41.700541ms 
RecursiveSNARK::prove_step 5: true, took 41.201209ms 
RecursiveSNARK::prove_step 6: true, took 41.845625ms 
RecursiveSNARK::prove_step 7: true, took 41.882541ms 
RecursiveSNARK::prove_step 8: true, took 41.978542ms 
RecursiveSNARK::prove_step 9: true, took 42.167542ms 
Verifying a RecursiveSNARK...
RecursiveSNARK::verify: true, took 28.778584ms
Generating a CompressedSNARK using Spartan with IPA-PC...
CompressedSNARK::prove: true, took 729.643958ms
CompressedSNARK::len 9406 bytes
Verifying a CompressedSNARK...
CompressedSNARK::verify: true, took 29.569208ms
=========================================================
```

### DIFF
--- a/src/bellpepper/r1cs.rs
+++ b/src/bellpepper/r1cs.rs
@@ -5,7 +5,9 @@
 use super::{shape_cs::ShapeCS, solver::SatisfyingAssignment, test_shape_cs::TestShapeCS};
 use crate::{
   errors::NovaError,
-  r1cs::{commitment_key, CommitmentKeyHint, R1CSInstance, R1CSShape, R1CSWitness},
+  r1cs::{
+    commitment_key, sparse::SparseMatrix, CommitmentKeyHint, R1CSInstance, R1CSShape, R1CSWitness,
+  },
   traits::Group,
   CommitmentKey,
 };
@@ -65,17 +67,15 @@ macro_rules! impl_nova_shape {
       G::Scalar: PrimeField,
     {
       fn r1cs_shape(&self) -> R1CSShape<G> {
-        let mut A: Vec<(usize, usize, G::Scalar)> = Vec::new();
-        let mut B: Vec<(usize, usize, G::Scalar)> = Vec::new();
-        let mut C: Vec<(usize, usize, G::Scalar)> = Vec::new();
+        let mut A = SparseMatrix::<G::Scalar>::empty();
+        let mut B = SparseMatrix::<G::Scalar>::empty();
+        let mut C: SparseMatrix<<G as Group>::Scalar> = SparseMatrix::<G::Scalar>::empty();
 
         let mut num_cons_added = 0;
         let mut X = (&mut A, &mut B, &mut C, &mut num_cons_added);
-
         let num_inputs = self.num_inputs();
         let num_constraints = self.num_constraints();
         let num_vars = self.num_aux();
-
         for constraint in self.constraints.iter() {
           add_constraint(
             &mut X,
@@ -85,29 +85,61 @@ macro_rules! impl_nova_shape {
             &constraint.2,
           );
         }
-
         assert_eq!(num_cons_added, num_constraints);
-
         let S: R1CSShape<G> = {
-          // Don't count One as an input for shape's purposes.
-          let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, &A, &B, &C);
+          let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, A, B, C);
           res.unwrap()
         };
-
         S
       }
     }
   };
 }
 
-impl_nova_shape!(ShapeCS);
+impl<G: Group> NovaShape<G> for ShapeCS<G>
+where
+  G::Scalar: PrimeField,
+{
+  fn r1cs_shape(&self) -> R1CSShape<G> {
+    let mut A = SparseMatrix::<G::Scalar>::empty();
+    let mut B = SparseMatrix::<G::Scalar>::empty();
+    let mut C: SparseMatrix<<G as Group>::Scalar> = SparseMatrix::<G::Scalar>::empty();
+    let mut num_cons_added = 0;
+    let mut X = (&mut A, &mut B, &mut C, &mut num_cons_added);
+
+    let num_inputs = self.num_inputs();
+    let num_constraints = self.num_constraints();
+    let num_vars = self.num_aux();
+
+    for constraint in self.constraints.iter() {
+      add_constraint(
+        &mut X,
+        num_vars,
+        &constraint.0,
+        &constraint.1,
+        &constraint.2,
+      );
+    }
+    assert_eq!(num_cons_added, num_constraints);
+
+    A.cols = num_vars + self.num_inputs();
+    B.cols = num_vars + self.num_inputs();
+    C.cols = num_vars + self.num_inputs();
+    
+    let S: R1CSShape<G> = {
+      let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, A, B, C);
+      res.unwrap()
+    };
+    S
+  }
+}
 impl_nova_shape!(TestShapeCS);
 
 fn add_constraint<S: PrimeField>(
   X: &mut (
-    &mut Vec<(usize, usize, S)>,
-    &mut Vec<(usize, usize, S)>,
-    &mut Vec<(usize, usize, S)>,
+    &mut SparseMatrix<S>,
+    &mut SparseMatrix<S>,
+    &mut SparseMatrix<S>,
     &mut usize,
   ),
   num_vars: usize,
@@ -117,29 +149,40 @@ fn add_constraint<S: PrimeField>(
 ) {
   let (A, B, C, nn) = X;
   let n = **nn;
-  let one = S::ONE;
+  assert_eq!(n + 1, A.indptr.len(), "A: invalid shape");
+  assert_eq!(n + 1, B.indptr.len(), "B: invalid shape");
+  assert_eq!(n + 1, C.indptr.len(), "C: invalid shape");
 
-  let add_constraint_component = |index: Index, coeff, V: &mut Vec<_>| {
+  let add_constraint_component = |index: Index, coeff: &S, M: &mut SparseMatrix<S>| {
     match index {
       Index::Input(idx) => {
         // Inputs come last, with input 0, reprsenting 'one',
         // at position num_vars within the witness vector.
-        let i = idx + num_vars;
-        V.push((n, i, one * coeff))
+        let idx = idx + num_vars;
+        M.data.push(*coeff);
+        M.indices.push(idx);
       }
-      Index::Aux(idx) => V.push((n, idx, one * coeff)),
+      Index::Aux(idx) => {
+        M.data.push(*coeff);
+        M.indices.push(idx);
+      }
     }
   };
 
   for (index, coeff) in a_lc.iter() {
     add_constraint_component(index.0, coeff, A);
   }
+  A.indptr.push(A.indices.len());
+
   for (index, coeff) in b_lc.iter() {
     add_constraint_component(index.0, coeff, B)
   }
+  B.indptr.push(B.indices.len());
+
   for (index, coeff) in c_lc.iter() {
     add_constraint_component(index.0, coeff, C)
   }
+  C.indptr.push(C.indices.len());
 
   **nn += 1;
 }
--- a/src/bellpepper/r1cs.rs
+++ b/src/bellpepper/r1cs.rs
@@ -76,6 +76,7 @@ macro_rules! impl_nova_shape {
         let num_inputs = self.num_inputs();
         let num_constraints = self.num_constraints();
         let num_vars = self.num_aux();
+
         for constraint in self.constraints.iter() {
           add_constraint(
             &mut X,
@@ -86,6 +87,11 @@ macro_rules! impl_nova_shape {
           );
         }
         assert_eq!(num_cons_added, num_constraints);
+
+        A.cols = num_vars + num_inputs;
+        B.cols = num_vars + num_inputs;
+        C.cols = num_vars + num_inputs;
+
         let S: R1CSShape<G> = {
           let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, A, B, C);
           res.unwrap()
@@ -125,7 +131,7 @@ where
     A.cols = num_vars + self.num_inputs();
     B.cols = num_vars + self.num_inputs();
     C.cols = num_vars + self.num_inputs();
-    
+
     let S: R1CSShape<G> = {
       let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, A, B, C);
       res.unwrap()
--- a/src/bellpepper/r1cs.rs
+++ b/src/bellpepper/r1cs.rs
@@ -92,53 +92,15 @@ macro_rules! impl_nova_shape {
         B.cols = num_vars + num_inputs;
         C.cols = num_vars + num_inputs;
 
-        let S: R1CSShape<G> = {
-          let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, A, B, C);
-          res.unwrap()
-        };
-        S
+        // Don't count One as an input for shape's purposes.
+        let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, A, B, C);
+        res.unwrap()
       }
     }
   };
 }
 
-impl<G: Group> NovaShape<G> for ShapeCS<G>
-where
-  G::Scalar: PrimeField,
-{
-  fn r1cs_shape(&self) -> R1CSShape<G> {
-    let mut A = SparseMatrix::<G::Scalar>::empty();
-    let mut B = SparseMatrix::<G::Scalar>::empty();
-    let mut C: SparseMatrix<<G as Group>::Scalar> = SparseMatrix::<G::Scalar>::empty();
-    let mut num_cons_added = 0;
-    let mut X = (&mut A, &mut B, &mut C, &mut num_cons_added);
-
-    let num_inputs = self.num_inputs();
-    let num_constraints = self.num_constraints();
-    let num_vars = self.num_aux();
-
-    for constraint in self.constraints.iter() {
-      add_constraint(
-        &mut X,
-        num_vars,
-        &constraint.0,
-        &constraint.1,
-        &constraint.2,
-      );
-    }
-    assert_eq!(num_cons_added, num_constraints);
-
-    A.cols = num_vars + self.num_inputs();
-    B.cols = num_vars + self.num_inputs();
-    C.cols = num_vars + self.num_inputs();
-
-    let S: R1CSShape<G> = {
-      let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, A, B, C);
-      res.unwrap()
-    };
-    S
-  }
-}
+impl_nova_shape!(ShapeCS);
 impl_nova_shape!(TestShapeCS);
 
 fn add_constraint<S: PrimeField>(
--- a/src/r1cs/sparse.rs
+++ b/src/r1cs/sparse.rs
@@ -171,7 +171,7 @@ impl<'a, F: PrimeField> Iterator for Iter<'a, F> {
     self.i += 1;
     // edge case at the end
     if self.i == self.nnz {
-      return Some(curr_item); // well this is ugly :(
+      return Some(curr_item);
     }
     // if `i` has moved to next row
     while self.i >= self.matrix.indptr[self.row + 1] {
--- a/src/r1cs/sparse.rs
+++ b/src/r1cs/sparse.rs
@@ -36,6 +36,16 @@ pub struct SparseMatrix<F: PrimeField> {
 }
 
 impl<F: PrimeField> SparseMatrix<F> {
+  /// 0x0 empty matrix
+  pub fn empty() -> Self {
+    SparseMatrix {
+      data: vec![],
+      indices: vec![],
+      indptr: vec![0],
+      cols: 0,
+    }
+  }
+
   /// construct from Vec<usize(row), usize(col), F>
   pub fn new(matrix: &[(usize, usize, F)], rows: usize, cols: usize) -> Self {
     let mut new_matrix = vec![vec![]; rows];
@@ -89,32 +99,12 @@ impl<F: PrimeField> SparseMatrix<F> {
       .collect()
   }
 
-  /// multiply by dense vector; should use rayon/gpu
-  /// we don't check dimensions
-  #[tracing::instrument(skip_all, name = "SparseMatrix::multiply_vec_unchecked")]
-  pub fn multiply_vec_unchecked(&self, vector: &[F]) -> Vec<F> {
-    // naive implementation for now
-    self
-      .indptr
-      .par_windows(2)
-      .map(|ptrs| {
-        let data = &self.data[ptrs[0]..ptrs[1]];
-        let indices = &self.indices[ptrs[0]..ptrs[1]];
-        data
-          .iter()
-          .zip(indices)
-          .map(|(val, col_idx)| *val * vector[*col_idx])
-          .sum()
-      })
-      .collect()
-  }
-
   /// number of non-zero entries
   pub fn len(&self) -> usize {
     *self.indptr.last().unwrap()
   }
 
-  /// number of non-zero entries
+  /// empty matrix
   pub fn is_empty(&self) -> bool {
     self.len() == 0
   }
--- a/src/r1cs/sparse.rs
+++ b/src/r1cs/sparse.rs
@@ -1,0 +1,259 @@
+//! # Sparse Matrices
+//!
+//! This module defines a custom implementation of CSR/CSC sparse matrices.
+//! Specifically, we implement sparse matrix / dense vector multiplication
+//! to compute the `A z`, `B z`, and `C z` in Nova.
+//!
+//! Notes:
+//!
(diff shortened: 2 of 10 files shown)


src/r1cs/sparse.rs
```diff
@@ -0,0 +1,259 @@
...
+      data.extend(val);
+    }
+
+    SparseMatrix {
+      data,
+      indices,
+      indptr,
+      cols,
+    }
+  }
```
huitseeker, line 69: This structure is effectively implementing a conversion from COO to CSR. The creation of the CSR matrices is done in the method `r1cs_shape`, in `impl NovaShape for MyGroup`, by re-constructing the matrix from constraints. 
Would the benchmark numbers be different if we implemented this in CSR form directly, without the detour through COO?

huitseeker (commented):
I think the core step here would be to not involve a COO -> CSR conversion but convert straight to CSR.

src/r1cs/sparse.rs
winston-h-zhang: Yes, I'll change this shortly; I was being lazy for the moment.

This conversion isn't in `NIFS::prove` or in any "critical path", so it's a one time cost.

winston-h-zhang:
Note: there are a few places where I don't do the most optimal calculation; I've neglected them in favor of getting this draft PR out. 

However, everything on the "critical path" is as one would theoretically expect. Maybe there are more concrete optimizations to be made.

porcuquine:
> This PR implements CSR matrix-vector multiplication under the `"csr"` feature. (This was for me to debug; if CSR sould become the only representation, then we can simply replace and remove the feature.) ~We get a slight uptick in performance, around 10%.~ Nevermind, it equals out as constraints get larger.

Given the motivation for this change, and that it will change digests, etc., I would love to see a stronger indication that the current implementation *will* produce the expected performance improvement. I fully accept that we have solid theoretical ground for expecting it, but it would be worth working out enough of the implementation to be sure we won't need more digest-affecting changes in order to realize the gains.

winston-h-zhang:
Here are some concrete gains in lurk-rs, bench `sha256_ivc_prove`, `rc=10, 100`.
No CSR:
```rust
sha256_ivc_prove/2023-09-05:7b1b685d5d39892eae74080ea179f2e1654a6f5e:rc=10:sha256_ivc_1/1                                                                          
                        time:   [9.3756 s 9.4571 s 9.5291 s]

sha256_ivc_prove/2023-09-05:7b1b685d5d39892eae74080ea179f2e1654a6f5e:rc=100:sha256_ivc_1/1                                                                          
                        time:   [6.2255 s 6.3031 s 6.3943 s]
```
CSR:
```rust
sha256_ivc_prove/2023-09-05:7b1b685d5d39892eae74080ea179f2e1654a6f5e:rc=10:sha256_ivc_1/1                                                                          
                        time:   [8.4676 s 8.6431 s 8.7572 s]
                        change: [-10.421% -8.6069% -7.0613%] (p = 0.00 < 0.05)
                        Performance has improved.
sha256_ivc_prove/2023-09-05:7b1b685d5d39892eae74080ea179f2e1654a6f5e:rc=100:sha256_ivc_1/1                                                                          
                        time:   [5.3944 s 5.5193 s 5.6346 s]
                        change: [-14.883% -12.435% -10.207%] (p = 0.00 < 0.05)
                        Performance has improved.
```

porcuquine:
> Here are some concrete gains in lurk-rs, bench `sha256_ivc_prove`, `rc=10, 100`. No CSR:

Ah, perfect — thanks. I think that's sufficient to show this is manifestly worthwhile. If we can gain confidence in the concrete representation being stable, I'd suggest we don't need a feature flag to accept this.

src/r1cs/sparse.rs
```diff
@@ -0,0 +1,259 @@
...
+        let data = &self.data[ptrs[0]..ptrs[1]];
+        let indices = &self.indices[ptrs[0]..ptrs[1]];
+        data
+          .iter()
+          .zip(indices)
+          .map(|(val, col_idx)| *val * vector[*col_idx])
+          .sum()
+      })
+      .collect()
+  }
```
huitseeker, line 110: I think this is overall not bad. The initial `assert_eq` check on the matrix length should be fine to keep.

huitseeker (commented):
Agreed with @porcuquine I think this is ripe for a CSR-only implementation, provided this uses CSR from the start. 

The conversion is indeed just part of setup, but when we're dealing with foldings that run in   <40ms, I expect the additional handful of ms involved in this unnecessary conversion to show a difference.

porcuquine:
> The conversion is indeed just part of setup, but when we're dealing with foldings that run in <40ms, I expect the additional handful of ms involved in this unnecessary conversion to show a difference.

It doesn't matter how long one folding takes. If we are folding many times, then the one-time cost of processing circuit shapes should not be a big deal. In other words, at the scale of proofs that will take seconds or minutes to generate, then ms are not that big a deal. However, I agree that we should just materialize whatever we are going to use, as a matter of efficiency in principal.

The reason I mention is that the only case (as I see it), where the one-time cost of creating the shape is significant, is the case where the whole proof is very very small. I'm trying to emphasize that this is not a very important case in terms of how we focus on optimization. So even though I agree it's worth this savings, the much more important point is to ensure that whatever we do here is indeed only being performed *once*, when creating parameters for a given circuit.

huitseeker:
> The reason I mention is that the only case (as I see it), where the one-time cost of creating the shape is significant, is the case where the whole proof is very very small. I'm trying to emphasize that this is not a very important case in terms of how we focus on optimization.

I agree, but that is very precisely the case chosen for the initial benchmarks @winston-h-zhang reported, which created some noise in communication.

porcuquine:
The logical implication of what I'm saying is that if we want to justify the work based on measured performance rather than theory, then we should choose a representative benchmark (like the ones from `lurk-rs` that I found convincing). What I was observing above is that if we have the expected performance characteristics, then on a suitable benchmark we *should* observe that this question of conversion at a cost on the order of milliseconds won't be significant. To be clear, I'm not arguing that we shouldn't favor the more efficient method given the total decision. I'm just saying that it is not a first-order performance consideration. If our benchmarks are forcing us to treat it as significant, we should either change them or be sophisticated enough to look past them before letting the tail wag the dog.

src/r1cs/sparse.rs
winston-h-zhang: `multiply_vec_unchecked` ended up getting removed

huitseeker (dismissed):
This work seems completely orthogonal to adding a digest to the R1CS shapes. In order to merge this quickly, I would recommend rebasing this PR so that it only contains the commits relevant to the CSR representation. There are two equivalent ways of doing this:
- `git rebase --onto` ([ref](https://womanonrails.com/git-rebase-onto))
- `git rebase --interactive origin/master` (and then remove all the irrelevant commits)

winston-h-zhang:
Yes, sorry the digest work was just for testing on lurk-rs. Let me rebase

huitseeker:
!benchmark

src/spartan/snark.rs
```diff
@@ -420,18 +421,14 @@ where
...
-        |M: &[(usize, usize, G::Scalar)], T_x: &[G::Scalar], T_y: &[G::Scalar]| -> G::Scalar {
-          (0..M.len())
-            .into_par_iter()
-            .map(|i| {
-              let (row, col, val) = M[i];
-              T_x[row] * T_y[col] * val
-            })
+        |M: &SparseMatrix<G::Scalar>, T_x: &[G::Scalar], T_y: &[G::Scalar]| -> G::Scalar {
+          M.iter()
+            .map(|(row, col, val)| T_x[row] * T_y[col] * val)
```
huitseeker, line 431: In the matrix-vector multiplication above, we let the inherent shape of the matrix drive the operation, by iterating over the matrix in a row-by-row manner (and then indexing into the vector's array using the column index).

It seems that here, rather than using the slow iterator, we could do the same thing: proceed row-by-row on the native CSR shape, indec into T_x, T_y using row & column indices.

src/spartan/snark.rs
```diff
@@ -182,9 +183,9 @@ where
         |S: &R1CSShape<G>, rx: &[G::Scalar]| -> (Vec<G::Scalar>, Vec<G::Scalar>, Vec<G::Scalar>) {
           assert_eq!(rx.len(), S.num_cons);
 
-          let inner = |M: &Vec<(usize, usize, G::Scalar)>, M_evals: &mut Vec<G::Scalar>| {
-            for (row, col, val) in M {
-              M_evals[*col] += rx[*row] * val;
+          let inner = |M: &SparseMatrix<G::Scalar>, M_evals: &mut Vec<G::Scalar>| {
+            for (row, col, val) in M.iter() {
+              M_evals[col] += rx[row] * val;
```
huitseeker, line 188: See my comment on `evaluate_with_table` below.

huitseeker (dismissed):
This needs a rebase, so back to your queue. 
I have a rebased branch, which I built for review, if that helps https://github.com/huitseeker/Nova/tree/csr-mvm-feature-review

Further, the iterator is OK when you really need to iterate over every element in a non-amiguous way[^1], e.g. when computing a hash. But it shouldn't be a crutch moving us away from using the native CSR shape for all arithmetic operations, which I think is possible here, in snark.rs and everything I've seen so far in ppsnark.rs.

[^1]:  i.e. where interpreting the order of arrival of the data must not require having context about the matrix representation

src/r1cs/sparse.rs
```diff
@@ -0,0 +1,249 @@
...
+//! to compute the `A z`, `B z`, and `C z` in Nova.
+//!
+//! Notes:
+//!
+//! There are many sparse matrix crates available in rust. However, most of
+//! these crates are do not parallelize, or are not compatiable with field elements.
+//!
+//! For example, [sprs](https://crates.io/crates/sprs/0.2.1) does not parallelize
+//! its sparse matrix / dense vector multiplication. It's traits also require
+//! `num_traits::Zero`, which is not compatiable with `ff::PrimeField`.
```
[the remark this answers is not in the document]
winston-h-zhang: 👍 Yes, I looked at that and I don't think there was parallelized implementations

src/bellpepper/r1cs.rs
```diff
@@ -85,29 +86,66 @@ macro_rules! impl_nova_shape {
...
         }
-
         assert_eq!(num_cons_added, num_constraints);
 
+        A.cols = num_vars + num_inputs;
+        B.cols = num_vars + num_inputs;
+        C.cols = num_vars + num_inputs;
+
         let S: R1CSShape<G> = {
-          // Don't count One as an input for shape's purposes.
```
huitseeker, line 92: This comment is load bearing for anyone that's not an R1CS expert so I'd like to keep it.

src/bellpepper/r1cs.rs
```diff
@@ -85,29 +86,66 @@ macro_rules! impl_nova_shape {
...
+      );
+    }
+    assert_eq!(num_cons_added, num_constraints);
+
+    A.cols = num_vars + self.num_inputs();
+    B.cols = num_vars + self.num_inputs();
+    C.cols = num_vars + self.num_inputs();
+
+    let S: R1CSShape<G> = {
+      let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, A, B, C);
```
huitseeker, line 136: Nit: is the intermediate variable useful?

src/bellpepper/r1cs.rs
```diff
@@ -85,29 +86,66 @@ macro_rules! impl_nova_shape {
...
         assert_eq!(num_cons_added, num_constraints);
 
+        A.cols = num_vars + num_inputs;
+        B.cols = num_vars + num_inputs;
+        C.cols = num_vars + num_inputs;
+
         let S: R1CSShape<G> = {
-          // Don't count One as an input for shape's purposes.
-          let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, &A, &B, &C);
+          let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, A, B, C);
```
huitseeker, line 96: Nit: is the intermediate variable useful?

src/bellpepper/r1cs.rs
```diff
@@ -85,29 +86,66 @@ macro_rules! impl_nova_shape {
...
         };
-
         S
       }
     }
   };
 }
 
-impl_nova_shape!(ShapeCS);
+impl<G: Group> NovaShape<G> for ShapeCS<G>
```
huitseeker, line 105: I may be missing something: could we fold this impl in a call of the `impl_nova_shape` macro?

src/r1cs/sparse.rs
```diff
@@ -0,0 +1,249 @@
...
+    SparseMatrix {
+      data: vec![],
+      indices: vec![],
+      indptr: vec![0],
+      cols: 0,
+    }
+  }
+
+  /// construct from Vec<usize(row), usize(col), F>
+  pub fn new(matrix: &[(usize, usize, F)], rows: usize, cols: usize) -> Self {
```
huitseeker, line 41: Is this needed for anything but unit testing? If not, can we assume at least that unit testing will provide matrices in sorted form, and then remove any correspondingly superfluous methods? If unit tests don't use this, would this be an invitation to?

src/r1cs/sparse.rs
```diff
@@ -0,0 +1,249 @@
...
+    Iter {
+      matrix: self,
+      row,
+      i: 0,
+      nnz: *self.indptr.last().unwrap(),
+    }
+  }
+
+  /// pad to specific dimensions, which must be bigger
+  pub fn pad(&mut self, rows: usize, num_vars_padded: usize, num_vars: usize) {
```
huitseeker, line 127: Could you add a unit test for the padding please?

src/spartan/snark.rs
```diff
@@ -195,9 +196,11 @@ where
         |S: &R1CSShape<G>, rx: &[G::Scalar]| -> (Vec<G::Scalar>, Vec<G::Scalar>, Vec<G::Scalar>) {
           assert_eq!(rx.len(), S.num_cons);
 
-          let inner = |M: &Vec<(usize, usize, G::Scalar)>, M_evals: &mut Vec<G::Scalar>| {
-            for (row, col, val) in M {
-              M_evals[*col] += rx[*row] * val;
+          let inner = |M: &SparseMatrix<G::Scalar>, M_evals: &mut Vec<G::Scalar>| {
+            for (row, ptrs) in M.indptr.windows(2).enumerate() {
```
huitseeker, line 200: Nit: a variable name such as row_idx instead of row would be a tad clearer.

src/spartan/snark.rs
```diff
@@ -433,17 +436,19 @@ where
...
-          (0..M.len())
-            .into_par_iter()
-            .map(|i| {
-              let (row, col, val) = M[i];
-              T_x[row] * T_y[col] * val
+        |M: &SparseMatrix<G::Scalar>, T_x: &[G::Scalar], T_y: &[G::Scalar]| -> G::Scalar {
+          M.indptr
+            .par_windows(2)
+            .enumerate()
+            .map(|(row, ptrs)| {
```
huitseeker, line 448: Same remark: `col_idx`, `row_idx` are generally nicer.

src/r1cs/sparse.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    SparseMatrix {
+      data,
+      indices,
+      indptr,
+      cols,
+    }
+  }
+
+  /// retrieves the data for row slice [i..j] from ptrs
```
huitseeker, line 72: It would be good if the comment could explain what is "unchecked" about this method.

huitseeker (commented):
This is generally quite good, a few missing items i commented on inline. Thanks!

src/r1cs/sparse.rs
winston-h-zhang: I've removed the sorting methods, but I've kept `::new()` for now, as it's used in `test_tiny_r1cs_with`. Also I've found that debugging is easier in COO form, so having a way to transform from COO to CSR via `::new()` and CSR to COO via `iter()` is nice.

src/r1cs/sparse.rs
huitseeker: The change in logic is completely fine, but I expect a unit test to show what's expected of this function. It's doubly needed if the internal logic is *not* what one would usually infer from the function's name.

src/r1cs/sparse.rs
huitseeker: Feel free to add the sorted assumption in `new`'s comment.

src/bellpepper/r1cs.rs
porcuquine: 💯 As I recall, Nova's R1CS shapes are not always intuitive, and any hints to let readers reconstruct the logic are worth retaining.

winston-h-zhang:
!benchmark

src/r1cs/sparse.rs
winston-h-zhang: Ok, I see what you mean. Would it make more sense to put the test in the `R1CSShape` module, though, since that is the central object being modified?

winston-h-zhang:
!benchmark

samuelburnham:
!benchmark

src/r1cs/sparse.rs
```diff
@@ -0,0 +1,228 @@
...
+    ];
+    let sparse_matrix = SparseMatrix::<Fr>::new(&matrix_data, 3, 3);
+    let vector = vec![Fr::from(1), Fr::from(2), Fr::from(3)];
+
+    let result = sparse_matrix.multiply_vec(&vector);
+
+    // Assuming the multiplication is done row by row,
+    // Result[0] = 2.0 * 2.0 = 4.0
+    // Result[1] = 3.0 * 3.0 = 9.0
+    // Result[2] = 4.0 * 1.0 = 4.0
```
huitseeker, line 210: This is nice, but I think it could be complemented with :
1. a test annotated with `#[should_panic]` testing that `multiply_vec` indeed checks the matrix-vector dimensions,
2. an example where there is more than one non-zero operand in a row's terms, to check the addition is performed correctly (at present this just tests multiplication). 
e.g. 
```rust
    let matrix_data = vec![
      (0, 1, Fr::from(2)),
      (0, 2, Fr::from(7)),
      (1, 2, Fr::from(3)),
      (2, 0, Fr::from(4)),
    ];
    let vector = vec![Fr::from(1), Fr::from(2), Fr::from(3)];
    // to be compared with vec![Fr::from(25), Fr::from(9), Fr::from(4)]
```

If you want to go the whole 9 yards, a proptest would be tip top, but not completely necessary.

huitseeker (commented):
I think this is close to done. The only blocking items are:
- the initialization of the inner vector in `fn new`,
- adjusting `test_matrix_vector_multiplication` and `test_matrix_iter` in any of the ways I suggested, or equivalent.
I've left a range of suggestions for those tests, but I'll be happy with a minimal fix.

src/bellpepper/r1cs.rs
```diff
@@ -85,16 +86,15 @@ macro_rules! impl_nova_shape {
...
-          // Don't count One as an input for shape's purposes.
-          let res = R1CSShape::new(num_constraints, num_vars, num_inputs - 1, &A, &B, &C);
-          res.unwrap()
-        };
+        A.cols = num_vars + num_inputs;
+        B.cols = num_vars + num_inputs;
+        C.cols = num_vars + num_inputs;
 
-        S
+        // Don't count One as an input for shape's purposes.
```
huitseeker, line 95: Thanks!

src/r1cs/sparse.rs
```diff
@@ -0,0 +1,228 @@
...
+      indices: vec![],
+      indptr: vec![0],
+      cols: 0,
+    }
+  }
+
+  /// Construct from the COO representation; Vec<usize(row), usize(col), F>.
+  /// We implicitly sort the columns during construction.
+  pub fn new(matrix: &[(usize, usize, F)], rows: usize, cols: usize) -> Self {
+    let mut new_matrix = vec![vec![]; rows];
```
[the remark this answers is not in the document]
winston-h-zhang: I don't think this can be done, because the length of the rows is unknown. Of course we know that a capacity of `cols` is an upper bound, but then we would be allocating `O(rows * cols)` memory upfront. In practice, if the matrix is actually sparse then the rows will be small and the cost of resizing should be minimal.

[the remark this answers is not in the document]
huitseeker: Oh, my apologies, my brain farted, and I forgot the semantics of the `rows` and `cols` arguments for a minute. my bad.

src/r1cs/sparse.rs
```diff
@@ -204,25 +212,46 @@ mod tests {
...
+    type Parameters = ();
+    type Strategy = BoxedStrategy<Self>;
+
+    fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+      let strategy = any::<[u8; 32]>()
+        .prop_map(|seed| FWrap(F::random(StdRng::from_seed(seed))))
+        .no_shrink();
+      strategy.boxed()
+    }
+  }
```
huitseeker, line 237: Oh, `FWrap` indeed only exists in Lurk. I think it's ok to push this in a utility module for now. Perhaps `src/r1cs/util.rs`?

src/r1cs/sparse.rs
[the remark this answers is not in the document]
winston-h-zhang: AFAIU, `rows` (represented by `indptr`) is fixed to be the actual number of rows. I guess we could also do dynamic row sizing, but that would be non-conventional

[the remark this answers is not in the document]
huitseeker: Fair enough, that makes sense

huitseeker (dismissed):

huitseeker (approved):
Tip top! thank you so much!

PR #39, opened 2023-09-06
Add benchmark comment action

samuelburnham:
Copy of the action in [lurk-rs](https://github.com/lurk-lab/lurk-rs/blob/a574b5198947b4d0dea80e349807c6222ca5a157/.github/workflows/bench_pr_comment.yml) with a self-hosted runner, also triggered via the `!benchmark` keyword in a PR comment. I wasn't sure which benchmark we wanted to run so I put `sha256` for now.

### DIFF
--- a/.github/workflows/bench_pr_comment.yml
+++ b/.github/workflows/bench_pr_comment.yml
@@ -1,0 +1,40 @@
+# Creates a PR benchmark comment with a comparison to main
+name: Benchmark pull requests
+on:
+  issue_comment:
+    types: [created]
+
+env:
+  CARGO_TERM_COLOR: always
+
+concurrency:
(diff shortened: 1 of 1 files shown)


huitseeker (commented):
> I wasn't sure which benchmark we wanted to run so I put sha256 for now.
Oh, there's an optimal answer to this:
- we always run `--bench recursive-snark`
- we run `--bench compressed-snark` if the PR touches anything in `src/spartan`,
- we run `--bench recursive-snark-supernova` if the PR touches anything in `src/supernova`

huitseeker (approved):
This LGTM, nice way to mutualize probing for changes. I'm happy with this not being perfect on the first try, we can iterate!

Issue #29, opened 2023-08-27
Avoid copies of the public parameters in Supernova's Running claims


huitseeker:
It seems quite essential to make sure that the `params` field of `RunningClaim` is an `Rc` to a `PublicParams` singleton, this sort of duplication is quickly going to be unmanageable.

Note the only reason I could see that we're manipulating a `Vec<RunningClaim<...>>` is to be able to index into them. We could have a `struct RunningClaim<...>` which would host the public parameter reference and some `impl Index`, `impl IndexMut` impls. Shall we open an issue?

_Originally posted by @huitseeker in https://github.com/lurk-lab/arecibo/pull/23#discussion_r1306699094_

porcuquine:
Looking into this, we do actually need distinct `PublicParams` per claim, since the shape is part of the params. The commitment keys could be shared.

porcuquine:
However, it's not clear this would be straightforward.

huitseeker:
That's fair, but instances of `PublicParams` only differentiated based on the `R1CSShape`, which is used in the fields `r1cs_shape_primary` and `r1cs_shape_secondary`. My claim is that, aside for those two fields (and to a lesser degree those that depend on them, as the `CommitmentKey`s are dependent on the max among the k-R1CSShapes in a k-IVC shape), everything else is the same between two instances of `PublicParams`. 

If we run n steps, each chosen among k circuit pairs in our k-IVC setup, we are keeping n copies of each field, whereas the minimum is at most k copies for the fields that depend strongly on the shape (the primary and secondary shapes themselves), plus 1 copy of everything else.

What I would therefore suggest for the purpose of setting this up is replacing the non-shape fields by a reference into a singleton, which could be implemented as an `Arc`. The setup of the `CommitmentKey` may be a bit more subtle since it requires updating references to it in n copies of the `RunningInstances` once it is set up, but an [`ArcSwapOption`](https://docs.rs/arc-swap/latest/arc_swap/) comes to mind as a way of doing this.

porcuquine:
Where do the n copies come from? I think there are just k copies, one held in each claim. New copies of the params are not made at each folding step, right?

PR #40, opened 2023-09-06
Add DigestBuilder.

porcuquine:
This PR implements a `DigestBuilder` and supplementary traits:
- `HasDigest`
- `Digestible`
- `SimpleDigestible`

Some elements of this PR are intended to support #31 (see #34) and #29. The goal is to begin work on a framework that will rationalize digest creation, at least across existing use cases and those in the linked issues. The present work moves us in that direction while maintaining the status quo by not affecting the values of existing digests.

Some details of the requirements implied by #29 and #31 will likely require further refactoring, but that will best be done in the context of that work rather than here. I mention this mainly to explain that the current PR is still somewhat rough and has some loose ends that will either need to be tightened up or removed later.

### DIFF
--- a/src/digest.rs
+++ b/src/digest.rs
@@ -90,9 +90,6 @@ impl<F: PrimeField, T: HasDigest<F> + Digestible> DigestBuilder<F, T> {
     let mut bytes: [u8; 32] = hasher.finalize().into();
     self.inner.set_digest(Self::map_to_field(&mut bytes));
 
-    // let digest = self.compute_digest(&self.inner)?;
-
-    // self.inner.set_digest(digest);
     Ok(self.inner)
   }
 }
--- a/src/digest.rs
+++ b/src/digest.rs
@@ -1,0 +1,119 @@
+use ff::PrimeField;
+use serde::Serialize;
+use sha3::{Digest, Sha3_256};
+use std::iter::Extend;
+use std::marker::PhantomData;
+
+use crate::constants::NUM_HASH_BITS;
+
+/// For building digests
+#[derive(Clone)]
+pub struct DigestBuilder<F: PrimeField, T: HasDigest<F>> {
+  bytes: Vec<u8>,
+  hasher: Option<Sha3_256>,
+  bytes_digest: [u8; 32],
+  inner: Option<T>,
+  _p: PhantomData<(F, T)>,
+}
+
+/// Trait to be implemented by types whose digests can be built with `DigestBuilder`.
+pub trait HasDigest<F: PrimeField> {
+  /// Extend `bytes` with raw bytes or digest summarizing `Digestible`.
+  fn set_digest(&mut self, digest: F);
+  fn set_bytes_digest(&mut self, _bytes_digest: [u8; 32]) {
+    unimplemented!()
+  }
+}
+
+/// Trait for components with potentially discrete digests to be included in their container's digest.
+pub trait Digestible {
+  /// Extend a byte sink with the bytes to be digested.
+  fn extend_bytes<X: Extend<u8>>(&self, bytes: &mut X);
+}
+
+/// Marker trait to be implemented for types that implement `Digestible` and `Serialize`.
+/// Their instances will be serialized to bytes then digested.
+pub trait SimpleDigestible: Serialize {}
+
+impl<T: SimpleDigestible> Digestible for T {
+  fn extend_bytes<X: Extend<u8>>(&self, bytes: &mut X) {
+    (*bytes).extend(bincode::serialize(self).unwrap());
+  }
+}
+
+impl<F: PrimeField, T: HasDigest<F> + Digestible> DigestBuilder<F, T> {
+  /// Return a new `DigestBuilder`.
+  pub fn new() -> Self {
+    Self {
+      bytes: Default::default(),
+      hasher: Some(Sha3_256::new()),
+      bytes_digest: [0; 32],
+      inner: None,
+      _p: Default::default(),
+    }
+  }
+
+  /// Initialize `DigestBuilder` with `inner`.
+  pub fn init(&mut self, inner: T) {
+    self.inner = Some(inner)
+  }
+
+  /// Update builder's hasher.
+  pub fn update(&mut self) {
+    if let Some(hasher) = self.hasher.as_mut() {
+      hasher.update(&self.bytes);
+    } else {
+      panic!("hasher missing");
+    };
+
+    self.bytes.truncate(0);
+  }
+
+  pub fn finalize_bytes_digest(&mut self) {
+    self.update();
+
+    let Some(hasher) = self.hasher.take() else {panic!("hasher missing");};
+    self.bytes_digest = hasher.finalize().into();
+  }
+
+  /// Finalize digest.
+  pub fn finalize(&mut self) -> F {
+    self.finalize_bytes_digest();
+
+    let digest = self.bytes_digest;
+    let bv = (0..NUM_HASH_BITS).map(|i| {
+      let (byte_pos, bit_pos) = (i / 8, i % 8);
+      let bit = (digest[byte_pos] >> bit_pos) & 1;
+      bit == 1
+    });
+
+    // turn the bit vector into a scalar
+    let mut digest = F::ZERO;
+    let mut coeff = F::ONE;
+    for bit in bv {
+      if bit {
+        digest += coeff;
+      }
+      coeff += coeff;
+    }
+    digest
+  }
+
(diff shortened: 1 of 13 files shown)


src/supernova/mod.rs
```diff
@@ -169,6 +177,108 @@ where
...
+  G1: Group<Base = <G2 as Group>::Scalar>,
+  G2: Group<Base = <G1 as Group>::Scalar>,
+  C1: StepCircuit<G1::Scalar>,
+  C2: StepCircuit<G2::Scalar>,
+{
+  /// Indexed `RunningClaim`s
+  pub claims: Vec<RunningClaim<G1, G2, C1, C2>>,
+  /// Digest constructed from these `RunningClaim`s' parameters
+  pub digest: G1::Scalar,
+}
```
hero78119, line 192: a nitpick: looks like can make them private scope with getter?

porcuquine: done

src/digest.rs
```diff
@@ -0,0 +1,119 @@
...
+      }
+      coeff += coeff;
+    }
+    digest
+  }
+
+  /// Extend `DigestBuilder` with the bytes provided by the underlying inner `HasDigest`.
+  pub fn compute_digest(&mut self) -> F {
+    let o = self.inner.as_ref().expect("inner Digestible missing");
+    o.extend_bytes(&mut self.bytes);
```
winston-h-zhang, line 105: It seems we write into `self.bytes` as a buffer and then immediately consume it with `finalize()`. However, there's no other place where `self.bytes` can be extended other than `new()`, which initializes it as empty. Can't this just be replaced locally with:
```rust
let mut bytes = vec![];
o.extend_bytes(&mut bytes);
```
Is `self.bytes` designing for future use case where we are extending `bytes` and updating the hasher dynamically?

src/digest.rs
```diff
@@ -0,0 +1,119 @@
...
+use std::iter::Extend;
+use std::marker::PhantomData;
+
+use crate::constants::NUM_HASH_BITS;
+
+/// For building digests
+#[derive(Clone)]
+pub struct DigestBuilder<F: PrimeField, T: HasDigest<F>> {
+  bytes: Vec<u8>,
+  hasher: Option<Sha3_256>,
```
winston-h-zhang, line 13: What is the reasoning behind this `Option<_>`?

src/digest.rs
porcuquine: Yes, the inner object may need to implement a more interesting (than just serializing itself) method of providing bytes. For example, if it contains other `Digestible` things (like circuits), then it will want to just provide its own digest — which may already have been computed. We don't necessarily need to actually accumulate all these bytes before hashing. Probably just updating the hasher itself will be fine (rather than saving the bytes and doing all the hashing last). I left it as-is to stay close to the previous work until we know exactly what the total process will be. It may be that changing to eliminate the accumulated bytes will have some effect on the digest itself because of potential padding considerations. My goal for now was to refactor without changing any digest results, though.

src/digest.rs
porcuquine: I believe it was because I could not finalize the hasher while it was still owned by the builder. The option lets me `take()` it before finalization.

src/digest.rs
```diff
@@ -0,0 +1,119 @@
...
+  /// Extend `DigestBuilder` with the bytes provided by the underlying inner `HasDigest`.
+  pub fn compute_digest(&mut self) -> F {
+    let o = self.inner.as_ref().expect("inner Digestible missing");
+    o.extend_bytes(&mut self.bytes);
+
+    self.finalize()
+  }
+
+  /// Build and return inner `Digestible`.
+  pub fn build(&mut self) -> T {
```
huitseeker, line 111: I think it's really important for `build` to consume `self`.

huitseeker (commented):
This `DigestBuilder` implementation seems to show a weird API that's distinct from the classics of the genre (see [e.g.](https://github.com/RustCrypto/traits/tree/master/digest)).

I suspect the following might work better:
https://gist.github.com/huitseeker/42995a133786b7a063b78e6a9bca86aa

porcuquine:
> This `DigestBuilder` implementation seems to show a weird API that's distinct from the classics of the genre (see [e.g.](https://github.com/RustCrypto/traits/tree/master/digest)).
> 
> I suspect the following might work better: https://gist.github.com/huitseeker/42995a133786b7a063b78e6a9bca86aa

That's quite likely. Much of what is here is the outcome of trying to understand and restructure exactly what already existed. Feel free to make a PR to this PR, if you think accepting it would address your comments above. I am not attached to the current state and assume it will need to change before this work is complete anyway.

huitseeker:
@porcuquine  https://github.com/lurk-lab/arecibo/pull/42
Apologies, I have not had the time to propagate the simpler `DigestBuilder` to files in the PR. 😢

src/digest.rs
```diff
@@ -0,0 +1,119 @@
...
+    let digest = self.compute_digest();
+
+    if let Some(d) = self.inner.as_mut() {
+      d.set_digest(digest)
+    }
+    self.inner.take().expect("inner Digestible missing")
+  }
+}
```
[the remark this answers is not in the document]
porcuquine: The initial idea was that you might want to finalize the hash *instead* of building (in some use cases). In that case, I would not expect this test to pass but rather to represent an intended programmer error. However, as noted above, I don't think we need to deal with it that way. I'm fine with simplifying the whole mechanism rather than applying point fixes to symptoms, though.

src/digest.rs
porcuquine: I agree.

porcuquine:
> @porcuquine #42 Apologies, I have not had the time to propagate the simpler `DigestBuilder` to files in the PR. 😢

No worries. I can pick it up later, unless someone else does. I'm generally fine with taking your simpler code — which is closer to where I think this would have ended up eventually anyway. I think it can perhaps go a bit further, but even if true it can wait.

src/digest.rs
```diff
@@ -0,0 +1,98 @@
...
+  /// Build and return inner `Digestible`.
+  pub fn build(mut self) -> Result<T, io::Error> {
+    let mut hasher = Self::hasher();
+    self.inner.write_bytes(&mut hasher)?;
+    let mut bytes: [u8; 32] = hasher.finalize().into();
+    self.inner.set_digest(Self::map_to_field(&mut bytes));
+
+    // let digest = self.compute_digest(&self.inner)?;
+
+    // self.inner.set_digest(digest);
```
huitseeker, line 95: We probably want to remove these.

src/lib.rs
```diff
@@ -46,59 +49,23 @@ use r1cs::{
...
-where
-  G1: Group<Base = <G2 as Group>::Scalar>,
-  G2: Group<Base = <G1 as Group>::Scalar>,
-  C1: StepCircuit<G1::Scalar>,
-  C2: StepCircuit<G2::Scalar>,
-  <G1::Scalar as PrimeField>::Repr: Abomonation,
-  <G2::Scalar as PrimeField>::Repr: Abomonation,
-)]
-pub struct PublicParams<G1, G2, C1, C2>
+impl<G1, G2, C1, C2> DigestBuilder<G1::Scalar, PublicParams<G1, G2, C1, C2>>
```
huitseeker, line 61: If we want to get fancy, we can do 
```
pub type PublicParamsBuilder<G1, G2, C1, G2> = DigestBuilder<G1::Scalar, PublicParams<G1, G2, C1, C2>>;
```

And similar for the others.

src/lib.rs
```diff
@@ -864,36 +902,6 @@ type Commitment<G> = <<G as Group>::CE as CommitmentEngineTrait<G>>::Commitment;
...
-/// compute digest giving a collection of Serialize object
-pub fn compute_digest<G: Group, T: Serialize>(o: &[&T]) -> G::Scalar {
-  // obtain a vector of bytes representing public parameters
-  let mut bytes = vec![];
-  o.iter()
-    .for_each(|o| bytes.extend(bincode::serialize(*o).unwrap()));
-  // convert pp_bytes into a short digest
-  let mut hasher = Sha3_256::new();
-  hasher.update(&bytes);
-  let digest = hasher.finalize();
```
huitseeker, line 876: Ah, I see what the issue is, and it's actually an important one. We can and should definitely accommodate this for the Supernova pps, which is the only callsite in the code base where this is called with a slice of size > 1. But this can be done in a subsequent PR.

src/digest.rs
porcuquine: done

huitseeker (dismissed):
We should figure out nested Digests indeed, in another PR.bthanks for this one!

huitseeker (approved):

PR #41, opened 2023-09-07
Enforce program counter in SuperNova StepCircuit.

porcuquine:
This PR formalizes the requirement that the new program counter returned from `circuit_supernova::StepCircuit::synthesize()` is the one assigned to the circuit that is actually synthesized at each step.

The previous `TestROM` example actually did *not* follow this convention but rather used the program counter for its own purposes, as an index into a predetermined sequence of operations — then enforced the correctness of this ad-hoc indexing scheme in each circuit.

The current work enforces the constraint that `pc` has been correctly calculated and returned from each circuit, requiring that it is interpreted precisely as that circuit's assigned index in the `RunningClaim`s.

In order to make the existing tests pass, they were reworked to re-envision the pseudo-pc enforcement at the start of each circuit — to a calculation (morally $\varphi$) of the new `pc` to be returned at the 'end' of the circuit. Without this change, the existing tests fail, exactly as expected, since `pc` was a monotonically increasing value, rather than the index of the chosen next circuit.

The circuit necessarily grows by one constraint.

Since the goal is that for the SuperNova user (the application/protocol/proof developer), the enforcement is performed by a new, internal trait `circuit_supernova::EnforcingStepCircuit` — which is automatically implemented for types that implement `circuit_supernova::StepCircuit`.

### DIFF
--- a/src/supernova/test.rs
+++ b/src/supernova/test.rs
@@ -2,12 +2,12 @@ use crate::bellpepper::test_shape_cs::TestShapeCS;
 use crate::gadgets::utils::alloc_const;
 use crate::gadgets::utils::alloc_num_equals;
 use crate::gadgets::utils::conditionally_select;
-use crate::gadgets::utils::{add_allocated_num, alloc_one, alloc_zero};
+use crate::gadgets::utils::{alloc_one, alloc_zero};
 use crate::provider::poseidon::PoseidonConstantsCircuit;
 use crate::traits::circuit_supernova::{
   EnforcingStepCircuit, StepCircuit, TrivialSecondaryCircuit, TrivialTestCircuit,
 };
-use bellpepper::gadgets::boolean::Boolean;
+use bellpepper::gadgets::{boolean::Boolean, Assignment};
 use bellpepper_core::num::AllocatedNum;
 use bellpepper_core::{ConstraintSystem, LinearCombination, SynthesisError};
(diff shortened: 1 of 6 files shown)


hero78119:
LGTM! it's nice to formalize and align terminology of `pc` with Supernova paper, so it will be friendly for audience came from paper reading. Besides it also cover ROM use case properly in test case.

emmorais (approved):

src/supernova/test.rs
```diff
@@ -80,40 +82,83 @@ where
...
+  cs.enforce(
+    || "Check u_fold",
+    |lc| lc + a.get_variable() + b.get_variable(),
+    |lc| lc + CS::one(),
+    |lc| lc + c.get_variable(),
+  );
+  Ok(c)
+}
+
+fn next_rom_index_and_pc<F: PrimeField, CS: ConstraintSystem<F>>(
```
huitseeker, line 103: This would  benefit from a doctoring no?

huitseeker (approved):
Thanks a bunch!

src/supernova/test.rs
porcuquine: I'm not sure what you mean. The name? I'm not keen on getting too deep into design/engineering of the 'API' for the toy ROM example, since it's pretty contrived in its current state anyway. But I may also miss what you are suggesting.

PR #42, opened 2023-09-08
refactor: Refactor DigestBuilder

huitseeker:
- Refactored `src/digest.rs` to replace `Vec<u8>` storage with dedicated Write I/O.
- Removed optional `hasher` and introduced dedicated factory method.
- Reworked digest computation and mapping into separate functions.
- Merged build and digest computation to enhance coherence.
- Improved type safety with Result error propagation.

### DIFF
--- a/benches/compressed-snark.rs
+++ b/benches/compressed-snark.rs
@@ -71,7 +71,8 @@ fn bench_compressed_snark(c: &mut Criterion) {
       &c_secondary,
       Some(S1::commitment_key_floor()),
       Some(S2::commitment_key_floor()),
-    );
+    )
+    .unwrap();
 
     // Produce prover and verifier keys for CompressedSNARK
     let (pk, vk) = CompressedSNARK::<_, _, _, _, S1, S2>::setup(&pp).unwrap();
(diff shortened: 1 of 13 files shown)


porcuquine:
The one thing I don't like about this is that it requires materializing *all* the bytes, whereas I think the better solution would allow just hashing them. This would require that the hasher live in the `DigestBuilder` as previously. I agree this is generally cleaner, but at that cost.

porcuquine:
I've pushed a commit that gets most of the way. However, there still remains a final knot to untie, suggesting the refactor has not yet solved the 'build should consume' issue:
```
error[E0502]: cannot borrow `self` as mutable because it is also borrowed as immutable
  --> src/digest.rs:92:18
   |
92 |     let digest = self.compute_digest(&self.inner)?;
   |                  ^^^^^--------------^-----------^
   |                  |    |              |
   |                  |    |              immutable borrow occurs here
   |                  |    immutable borrow later used by call
   |                  mutable borrow occurs here
```

porcuquine:
Fixed now by eliminating the `compute_digest` method entirely.

porcuquine (approved):

PR #51, opened 2023-09-19
Setup SuperNova public parameter infrastructure

### LINKED ISSUE #29: Avoid copies of the public parameters in Supernova's Running claims

winston-h-zhang:
We split out the `PublicParams` within each `RunningClaim` into a new `supernova::PublicParams`. This split prepares for downstream changes in lurk-rs (https://github.com/lurk-lab/lurk-rs/pull/648) to working with the public parameter infra. 

- We aggregate the common primary commitment key and any other shared information amongst all the params into the main `PublicParams` structure and create a new `CircuitShape` structure to manage per-circuit parameters. 
- We mirror the Nova API and modify the supernova `prove`/`verify` functions to take a public params, as well as computing digests.
- We remove `RunningClaims`, which ends being a glorified `usize`.

Closes #29

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,6 +56,33 @@ use traits::{
   AbsorbInROTrait, Group, ROConstants, ROConstantsCircuit, ROTrait,
 };
 
+/// A type that holds parameters for the primary and secondary circuits of Nova and SuperNova
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Abomonation)]
+#[serde(bound = "")]
+#[abomonation_bounds(where <G::Scalar as PrimeField>::Repr: Abomonation)]
+pub struct CircuitShape<G: Group> {
+  F_arity: usize,
+  r1cs_shape: R1CSShape<G>,
+}
+
+impl<G: Group> SimpleDigestible for CircuitShape<G> {}
+
+impl<G: Group> CircuitShape<G> {
+  /// Create a new `CircuitShape`
+  pub fn new(r1cs_shape: R1CSShape<G>, F_arity: usize) -> Self {
+    Self {
+      F_arity,
+      r1cs_shape,
+    }
+  }
+
+  /// Return the [CircuitShape]' digest.
+  pub fn digest(&self) -> G::Scalar {
+    let dc: DigestComputer<'_, <G as Group>::Scalar, CircuitShape<G>> = DigestComputer::new(self);
+    dc.digest().expect("Failure in computing digest")
+  }
+}
+
 /// A type that holds public parameters of Nova
 #[derive(Clone, PartialEq, Serialize, Deserialize, Abomonation)]
 #[serde(bound = "")]
@@ -80,11 +107,11 @@ where
   ro_consts_primary: ROConstants<G1>,
   ro_consts_circuit_primary: ROConstantsCircuit<G2>,
   ck_primary: CommitmentKey<G1>,
-  r1cs_shape_primary: R1CSShape<G1>,
+  circuit_shape_primary: CircuitShape<G1>,
   ro_consts_secondary: ROConstants<G2>,
   ro_consts_circuit_secondary: ROConstantsCircuit<G1>,
   ck_secondary: CommitmentKey<G2>,
-  r1cs_shape_secondary: R1CSShape<G2>,
+  circuit_shape_secondary: CircuitShape<G2>,
   augmented_circuit_params_primary: NovaAugmentedCircuitParams,
   augmented_circuit_params_secondary: NovaAugmentedCircuitParams,
   #[abomonation_skip]
@@ -179,6 +206,7 @@ where
     let mut cs: ShapeCS<G1> = ShapeCS::new();
     let _ = circuit_primary.synthesize(&mut cs);
     let (r1cs_shape_primary, ck_primary) = cs.r1cs_shape_and_key(optfn1);
+    let circuit_shape_primary = CircuitShape::new(r1cs_shape_primary, F_arity_primary);
 
     // Initialize ck for the secondary
     let circuit_secondary: NovaAugmentedCircuit<'_, G1, C2> = NovaAugmentedCircuit::new(
@@ -190,6 +218,7 @@ where
     let mut cs: ShapeCS<G2> = ShapeCS::new();
     let _ = circuit_secondary.synthesize(&mut cs);
     let (r1cs_shape_secondary, ck_secondary) = cs.r1cs_shape_and_key(optfn2);
+    let circuit_shape_secondary = CircuitShape::new(r1cs_shape_secondary, F_arity_secondary);
 
     PublicParams {
       F_arity_primary,
@@ -197,11 +226,11 @@ where
       ro_consts_primary,
       ro_consts_circuit_primary,
       ck_primary,
-      r1cs_shape_primary,
+      circuit_shape_primary,
       ro_consts_secondary,
       ro_consts_circuit_secondary,
       ck_secondary,
-      r1cs_shape_secondary,
+      circuit_shape_secondary,
       augmented_circuit_params_primary,
       augmented_circuit_params_secondary,
       digest: OnceCell::new(),
@@ -221,16 +250,16 @@ where
   /// Returns the number of constraints in the primary and secondary circuits
   pub const fn num_constraints(&self) -> (usize, usize) {
     (
-      self.r1cs_shape_primary.num_cons,
-      self.r1cs_shape_secondary.num_cons,
+      self.circuit_shape_primary.r1cs_shape.num_cons,
+      self.circuit_shape_secondary.r1cs_shape.num_cons,
     )
   }
 
   /// Returns the number of variables in the primary and secondary circuits
   pub const fn num_variables(&self) -> (usize, usize) {
     (
-      self.r1cs_shape_primary.num_vars,
-      self.r1cs_shape_secondary.num_vars,
+      self.circuit_shape_primary.r1cs_shape.num_vars,
+      self.circuit_shape_secondary.r1cs_shape.num_vars,
     )
   }
 }
@@ -296,7 +325,7 @@ where
       .map_err(|_| NovaError::SynthesisError)
       .expect("Nova error synthesis");
     let (u_primary, w_primary) = cs_primary
-      .r1cs_instance_and_witness(&pp.r1cs_shape_primary, &pp.ck_primary)
+      .r1cs_instance_and_witness(&pp.circuit_shape_primary.r1cs_shape, &pp.ck_primary)
       .map_err(|_e| NovaError::UnSat)
       .expect("Nova error unsat");
 
@@ -322,23 +351,27 @@ where
       .map_err(|_| NovaError::SynthesisError)
       .expect("Nova error synthesis");
     let (u_secondary, w_secondary) = cs_secondary
-      .r1cs_instance_and_witness(&pp.r1cs_shape_secondary, &pp.ck_secondary)
+      .r1cs_instance_and_witness(&pp.circuit_shape_secondary.r1cs_shape, &pp.ck_secondary)
       .map_err(|_e| NovaError::UnSat)
       .expect("Nova error unsat");
 
     // IVC proof for the primary circuit
     let l_w_primary = w_primary;
     let l_u_primary = u_primary;
-    let r_W_primary = RelaxedR1CSWitness::from_r1cs_witness(&pp.r1cs_shape_primary, &l_w_primary);
-    let r_U_primary =
-      RelaxedR1CSInstance::from_r1cs_instance(&pp.ck_primary, &pp.r1cs_shape_primary, &l_u_primary);
+    let r_W_primary =
+      RelaxedR1CSWitness::from_r1cs_witness(&pp.circuit_shape_primary.r1cs_shape, &l_w_primary);
+    let r_U_primary = RelaxedR1CSInstance::from_r1cs_instance(
+      &pp.ck_primary,
+      &pp.circuit_shape_primary.r1cs_shape,
+      &l_u_primary,
+    );
 
     // IVC proof of the secondary circuit
     let l_w_secondary = w_secondary;
     let l_u_secondary = u_secondary;
-    let r_W_secondary = RelaxedR1CSWitness::<G2>::default(&pp.r1cs_shape_secondary);
+    let r_W_secondary = RelaxedR1CSWitness::<G2>::default(&pp.circuit_shape_secondary.r1cs_shape);
     let r_U_secondary =
-      RelaxedR1CSInstance::<G2>::default(&pp.ck_secondary, &pp.r1cs_shape_secondary);
+      RelaxedR1CSInstance::<G2>::default(&pp.ck_secondary, &pp.circuit_shape_secondary.r1cs_shape);
 
     assert!(
       !(zi_primary.len() != pp.F_arity_primary || zi_secondary.len() != pp.F_arity_secondary),
@@ -398,7 +431,7 @@ where
       &pp.ck_secondary,
       &pp.ro_consts_secondary,
       &scalar_as_base::<G1>(pp.digest()),
-      &pp.r1cs_shape_secondary,
+      &pp.circuit_shape_secondary.r1cs_shape,
       &self.r_U_secondary,
       &self.r_W_secondary,
       &self.l_u_secondary,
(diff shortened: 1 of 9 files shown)


src/supernova/mod.rs
```diff
@@ -176,92 +176,182 @@ where
...
-  /// Indexed `RunningClaim`s
-  claims: Vec<RunningClaim<G1, G2, C1, C2>>,
+  /// The internal public params
+  pub pp_vec: Vec<PublicParams<G1, G2>>,
   /// Digest constructed from these `RunningClaim`s' parameters
   // Once we serialize this struct, it's important to add this
   // annotaiton to the digest field for it to be treated properly: #[serde(skip, default = "OnceCell::new")]
   digest: OnceCell<G1::Scalar>,
+  stage: Stage,
+  _p: PhantomData<C1>,
```
huitseeker, line 193: Why is there no parametrization by C2?

src/supernova/mod.rs
```diff
@@ -176,92 +176,182 @@ where
...
-  C2: StepCircuit<G2::Scalar>,
+  C1: EnforcingStepCircuit<G1::Scalar>,
 {
-  fn new(mut claims: Vec<RunningClaim<G1, G2, C1, C2>>) -> Self {
-    let running_claim_params = claims
+  /// new running claim params
+  pub fn new<NC: NonUniformCircuit<G1, G2, C1>>(
+    non_unifrom_circuit: &NC,
+    // optfn1: Option<CommitmentKeyHint<G1>>,
+    // optfn2: Option<CommitmentKeyHint<G2>>,
```
huitseeker, line 238: This is not needed if you’re doing a setup without CK.

src/supernova/mod.rs
```diff
@@ -176,92 +176,182 @@ where
...
+
+    running_claim_params.complete();
+    running_claim_params
+  }
 
-    for claim in &mut claims {
-      claim.set_commitment_key(ck_primary.clone(), ck_secondary.clone());
+  /// Compute primary and secondary commitment keys sized to handle the largest of the circuits in the provided
+  /// `PublicParams`.
+  fn compute_commitment_keys(&self) -> (CommitmentKey<G1>, CommitmentKey<G2>) {
```
huitseeker, line 303: This is where the optfn1, optfn2 parameters from above come in. They need to be parametrized by a preference expressed as a static function of the SNARK that will eventually be used to compress the proof. See #5. We should open an issue to fix this across SuperNova.

src/supernova/mod.rs
```diff
@@ -176,92 +176,182 @@ where
...
-  /// Return the `RunningClaims`' digest.
+  /// set primary/secondary commitment key for all the params
+  fn set_commitment_keys(&mut self) {
+    if matches!(self.stage, Stage::Complete) {
+      return;
+    }
+    let (ck_primary, ck_secondary) = self.compute_commitment_keys();
+    for pp in self.pp_vec.iter_mut() {
+      pp.ck_primary = Some(ck_primary.clone()); // TODO: the keys should be shared across all the params
+      pp.ck_secondary = Some(ck_secondary.clone()); // TODO: the keys should be shared across all the params
```
huitseeker, line 349: Hint for why I think this deserves more thought:
- the CKs are ginormous,
- they are a universal parameter, i.e. they are not circuit-dependent. They depend on the size of the circuit (num of variables and constraints, and matrix size, see #5), but can be shared between any circuits that are under the max size that was passed at their creation. See #5.

src/supernova/mod.rs
```diff
@@ -965,3 +1059,32 @@ where
...
+  // Initialize ck for the primary
+  let augmented_circuit: SuperNovaAugmentedCircuit<'_, G2, C> = SuperNovaAugmentedCircuit::new(
+    &augmented_circuit_params,
+    None,
+    circuit,
+    ro_consts_circuit,
+    num_augmented_circuits,
+  );
+  let mut cs: ShapeCS<G1> = ShapeCS::new();
+  let _ = augmented_circuit.synthesize(&mut cs);
```
huitseeker, line 1088: Same question as `circuit_digest` in `src/lib.rs`

huitseeker (commented):
I think your intuition that the old `RunningClaim` is a shallow, and mostly unnecessary shell around a `Vec` of ` PublicParams` is mostly correct, but I left comments inline that make me think the current struct re-organization isn’t quite where we want to be. I’ve left comments inline, but overall:
- the new `RunningClaim` is even more shallow (it’s a glorified usize)
- it’s become hard to set the commitment keys, because they are copied in every `PublicParam`.

I think some principles that point the direction of a better place are:
- the `CK` is common to a whole proof,
- in the `PublicParams`, everything but the circuit shapes is mutualized,
- the `RunningClaims` only exist in their plural form, we have usize-indexed APIs that can `get` the required info out of them,
- in order to not have to revamp all the APIs, we pass structs (parametrized by a `’ a` lifetime) that embark references to shared elements rather than copies.

src/supernova/mod.rs
winston-h-zhang: Fixed. This was originally because `NonUniformCircuit` wasn't parametriation by `C2`, either.

src/supernova/mod.rs
winston-h-zhang: Done in #53

src/supernova/mod.rs
winston-h-zhang: I attempted this by moving all the shared stuff into the topmost struct. It's a bit confusing since I've renamed everything, so please take a look at the code.

src/lib.rs
```diff
@@ -875,6 +875,29 @@ where
...
+
+  // ro_consts_circuit are parameterized by G2 because the type alias uses G2::Base = G1::Scalar
+  let ro_consts_circuit: ROConstantsCircuit<G2> = ROConstantsCircuit::<G2>::default();
+
+  // Initialize ck for the primary
+  let augmented_circuit: NovaAugmentedCircuit<'_, G2, C> =
+    NovaAugmentedCircuit::new(&augmented_circuit_params, None, circuit, ro_consts_circuit);
+  let mut cs: ShapeCS<G1> = ShapeCS::new();
+  let _ = augmented_circuit.synthesize(&mut cs);
+  cs.r1cs_shape().digest()
```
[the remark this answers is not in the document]
winston-h-zhang: The underlying issue is that we must know the circuit shape in some form to create a valid cache key. Thus, performing `synthesize` is unavoidable (unless there is another way to get the concrete representation of a circuit).

However, as you say, this is expensive. @porcuquine and I discussed this issue at length in [this zulip thread](https://zulip.lurk-lab.com/#narrow/stream/47-systems/topic/Public.20Parameters) and our conclusion was to rely on an "ad-hoc" representation of `(circuit(rc=2), actual_rc)`, which is much faster. So we know by design that this call *should* never be expensive because it is only used [here](https://github.com/lurk-lab/lurk-rs/pull/648/files#diff-cefb89359ae9c92029a66a179ecbcd6f8dc74f04d4d0d77fef25c77f536744c6R598-R608) in lurk-rs. This is hackish but it seems to be the best solution for now.

src/supernova/mod.rs
```diff
@@ -176,92 +176,182 @@ where
...
+    running_claim_params.complete();
+    running_claim_params
+  }
+
+  /// Create a [RunningClaimParams] from a vector of raw [PublicParams].
+  /// We check the state of the commitment keys of the given params and
+  /// realign them if necessary. If a digest is given, we assume that it
+  /// will correctly match the public params after realigning them, and
+  /// we do not check for its validity.
+  pub fn from_pp_vec(pp_vec: Vec<PublicParams<G1, G2>>, digest: OnceCell<G1::Scalar>) -> Self {
```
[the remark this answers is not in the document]
winston-h-zhang: I'm consuming the vector so it has to be allocated

src/lib.rs
```diff
@@ -875,6 +875,29 @@ where
...
+pub fn circuit_digest<
+  G1: Group<Base = <G2 as Group>::Scalar>,
+  G2: Group<Base = <G1 as Group>::Scalar>,
+  C: StepCircuit<G1::Scalar>,
+>(
+  circuit: &C,
+  is_primary_circuit: bool,
+) -> G1::Scalar {
+  let augmented_circuit_params =
+    NovaAugmentedCircuitParams::new(BN_LIMB_WIDTH, BN_N_LIMBS, is_primary_circuit);
```
winston-h-zhang, line 888: > the `is_primary_circuit` flag makes a ton of sense when you define a generic struct parameterized by a group `G` that can be used symmetrically to speak of `G1` or `G2`. In this case, the flag allows you to detect when that generic structure is being called with the primary group of the curve, and introduce an asymmetry of behavior within the generic function, for that case.
> 
> There’s probably a simpler way to proceed for the present function, isn’t there?

I don't see how; here we need the flag to tell us how to construct the circuit params

winston-h-zhang:
@huitseeker The biggest changes from your suggestions was getting rid of `RunningClaims` and just using a `usize`, renaming `RunningClaimParams` to `PublicParams` and `PublicParams` to `CircuitParams`.

src/lib.rs
huitseeker: Well, it seems that when we're considering a group pair, `G1` is by convention the primary. And the present function can't be called on anything but G1. 

But this is a nit, if your point is the function could theoretically be called with typed reversed, fine. Note we already hardcode this in the constructor of `PublicParams` though.

src/supernova/mod.rs
```diff
@@ -129,139 +113,226 @@ where
...
-      self.r1cs_shape_primary.num_vars,
-      self.r1cs_shape_secondary.num_vars,
-    )
-  }
 }
 
-/// SuperNova takes Ui a list of running instances.
-/// One instance of Ui is a struct called RunningClaim.
-pub struct RunningClaim<G1, G2, C1, C2>
+/// A vector of [CircuitParams] corresponding to a set of [RunningClaims]
```
huitseeker, line 126: The comments could use a pass of renaming.

src/supernova/mod.rs
```diff
@@ -129,139 +113,226 @@ where
...
 }
 
-impl<G1, G2, C1, C2> Index<usize> for RunningClaims<G1, G2, C1, C2>
+impl<G1, G2, C1, C2> Index<usize> for PublicParams<G1, G2, C1, C2>
 where
   G1: Group<Base = <G2 as Group>::Scalar>,
   G2: Group<Base = <G1 as Group>::Scalar>,
-  C1: StepCircuit<G1::Scalar>,
-  C2: StepCircuit<G2::Scalar>,
+  C1: EnforcingStepCircuit<G1::Scalar>,
```
huitseeker, line 187: This extension of bounds to an `EnforcingStepCircuit` should not be needed, specially if you don't use `enforcing_synthesis`. 

The way this works is when using the method `synthesize` on any `SuperNovaAugmentedCircuit`, Rust will find there is a requirement of the circuit being an `EnforcingStepCircuit`, and seeking that instance, will find the blanket implementation of this [extension trait](http://xion.io/post/code/rust-extension-traits.html):
https://github.com/lurk-lab/arecibo/blob/874c06c0203a0977cd3fc2f5848ca9649a1b988f/src/supernova/circuit.rs#L422-L423

Promoting requirements to an `EnforcingStepCircuit` is just noise. Consider the following diff please:
https://gist.github.com/huitseeker/74a4c4ee55e8b15459265236b6703c9e

src/supernova/mod.rs
```diff
@@ -129,139 +113,226 @@ where
...
+      ro_consts_secondary,
+      ro_consts_circuit_secondary,
+      ck_secondary,
+      r1cs_shape_secondary,
+      augmented_circuit_params_secondary,
       digest: OnceCell::new(),
+      _p: PhantomData,
+    };
+
+    pp.digest(); // make sure to compute the digest
```
huitseeker, line 269: Please clarify why in the comment. We wouldn't like folks to cargo-cult this call.

src/supernova/mod.rs
```diff
@@ -129,139 +113,226 @@ where
...
+  C1: EnforcingStepCircuit<G1::Scalar>,
+  C2: EnforcingStepCircuit<G2::Scalar>,
 {
-  fn new(mut claims: Vec<RunningClaim<G1, G2, C1, C2>>) -> Self {
-    let running_claim_params = claims
+  /// new running claim params
+  pub fn new<NC: NonUniformCircuit<G1, G2, C1, C2>>(
+    non_unifrom_circuit: &NC,
+  ) -> Self {
+    // uhhh if this aint zero this doesn't work
```
huitseeker, line 224: Does the output value of `iter_base_step` clear things up, and if so, could we refer to it here?

src/lib.rs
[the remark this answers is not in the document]
huitseeker: Please add a doc comment to this function and its supernova alter ego to state that this call is quite expensive, then.

huitseeker (commented):
Would it be possible to adapt `benches/recursive-snark-supernova.rs`?

src/supernova/mod.rs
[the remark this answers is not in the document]
huitseeker: This is obsolete given the new `from_parts`, but the idea is to avoid the allocation of the vector that precedes this call.

src/supernova/mod.rs
```diff
@@ -60,208 +116,205 @@ where
...
+      ro_consts_circuit_primary: aux_params.ro_consts_circuit_primary,
+      ck_primary: aux_params.ck_primary,
+      augmented_circuit_params_primary: aux_params.augmented_circuit_params_primary,
+      F_arity_secondary: aux_params.F_arity_secondary,
+      ro_consts_secondary: aux_params.ro_consts_secondary,
+      ro_consts_circuit_secondary: aux_params.ro_consts_circuit_secondary,
+      ck_secondary: aux_params.ck_secondary,
+      r1cs_shape_secondary: aux_params.r1cs_shape_secondary,
+      augmented_circuit_params_secondary: aux_params.augmented_circuit_params_secondary,
+      digest: aux_params.digest.into(),
```
huitseeker, line 300: The preferred way to do this would be to not trust the `AuxParams` digest and simply re-compute the digest, and compare it to the digest of the `AuxParams`. 
i.e. `assert_eq!(self.digest() == aux_params.digest)` which guarantees the correctness of the digest and that the user hasn't matched up the wrong set of circuits.
What you've implemented should pehaps be called `from_parts_unchecked` instead?

huitseeker (commented):
Just one remark left, otherwise, this LGTM! You may want to mark that this closes #29

src/supernova/test.rs
```diff
@@ -442,8 +443,8 @@ where
   let test_rom = TestROM::<G1, G2, TrivialSecondaryCircuit<G2::Scalar>>::new(rom);
   let num_steps = test_rom.num_steps();
 
-  let running_claim_params = RunningClaimParams::new(&test_rom);
-  let running_claims = test_rom.setup_running_claims(&running_claim_params);
+  let pp = PublicParams::new(&test_rom);
+  // let running_claims = test_rom.setup_running_claims(&running_claim_params);
```
huitseeker, line 447: Nit: this would need cleaning up.

src/supernova/mod.rs
```diff
@@ -51,307 +51,299 @@ pub(crate) mod utils;
...
+
+  /// Digest constructed from this `PublicParams`' parameters
+  // Once we serialize this struct, it's important to add this
+  // annotaiton to the digest field for it to be treated properly: #[serde(skip, default = "OnceCell::new")]
+  digest: OnceCell<G1::Scalar>,
+  _p: PhantomData<(C1, C2)>,
 }
 
-impl<G1, G2> SimpleDigestible for PublicParams<G1, G2>
+/// Auxillary [PublicParams] information about the commitment keys and
```
huitseeker, line 107: Nit: "Auxiliary"

huitseeker (commented):
This LGTM, but the current version of Lurk would break ([gist](https://gist.github.com/huitseeker/7ee3150fab4aebe15850dd00e4694646)) if we were to merge this, so I'll wait until https://github.com/lurk-lab/lurk-rs/pull/648 is ready to stamp here.

huitseeker:
@winston-h-zhang this will need a minor rebase due to #56

huitseeker:
This needs an also-minor rebase due to #58 (`TrivialTestCircuit` is renamed to `TrivialCircuit`)

winston-h-zhang:
@huitseeker I've changed `CircuitParams -> CircuitShape` and refactored it to be consistent among Nova/SuperNova.

huitseeker (approved):
Thanks a bunch!

arthurpaulino:
Is this a breaking change?

huitseeker:
> Is this a breaking change?

Yes.

PR #52, opened 2023-09-25
feat: non-deterministic NIVC example

mpenciak:
This PR adds a pair of NIVC examples.

The first, `nivc_static_phi`, incorporates non-deterministic hints from the prover to prove the chain of computations of cube and fifth roots.

The second, `nivc_dynamic_phi`, calculates the `program_counter` dynamically at each step as the least-significant bit of the calculated hash (either `sha256` or `blake2s`).


examples/nivc_dynamic_phi.rs
```diff
@@ -0,0 +1,360 @@
...
+      })
+      .collect::<Vec<Boolean>>();
+
+    let digest = pack_bits(cs.namespace(|| "digest_from_bits"), digest_bits.as_slice())?;
+
+    let new_pc = AllocatedNum::alloc(cs.namespace(|| "next_pc"), || {
+      if self.next_pc {
+        Ok(F::ONE)
+      } else {
+        Ok(F::ZERO)
```
porcuquine, line 85: What purpose does `self.next_pc` serve here? Why not just look at the value of `next_pc_bit`?

porcuquine: In other words, why do you need any hint in the `ShaCircuit` struct? I don't think you do.

examples/nivc_dynamic_phi.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    let preimage_bits = preimage_bits
+      .chunks(8)
+      .flat_map(|chunk| {
+        let mut chunk = chunk.to_vec();
+        chunk.reverse();
+        chunk
+      })
+      .collect::<Vec<_>>();
+
```
porcuquine, line 63: I think you could just pack the 80 bits you want into the preimage here — same as you do for the digest below. That would save the need for the strict decomposition and to constrain the unused bits (so savings of 200+ constraints).

porcuquine: This only works if the inputs will always be 80 bits, but I think that is the case for all the internal inputs, if all of your outputs are 80 bits (stuffed into a field element). So if you similarly restrict the initial input, then this would work.

examples/nivc_dynamic_phi.rs
```diff
@@ -0,0 +1,360 @@
...
+#[derive(Clone, Debug)]
+struct BlakeCircuit<F: PrimeField> {
+  next_pc: bool,
+  _p: PhantomData<F>,
+}
+
+impl<F: PrimeField> BlakeCircuit<F> {
+  fn new(next_pc: bool) -> Self {
+    Self {
+      next_pc,
```
porcuquine, line 109: Same comments/questions about the need for this hint as in the sha circuit above.

examples/nivc_dynamic_phi.rs
```diff
@@ -0,0 +1,360 @@
...
+
+#[derive(Debug)]
+struct ExampleSteps<G1, G2>
+where
+  G1: Group<Base = <G2 as Group>::Scalar>,
+  G2: Group<Base = <G1 as Group>::Scalar>,
+{
+  next_pc: bool,
+  _p: PhantomData<(G1, G2)>,
+}
```
porcuquine, line 166: I don't understand the need for or idea behind the struct. The name implies a collection of steps, yet you're only representing a single bit of information. Moreover, this is 'redundant hint' as described above. And as far as I can tell, this is a second level of indirection that redundant hint — since you're also keeping such hints in the individual hash circuits.

porcuquine:
I think the primary need this worked toward addressing is fulfilled by #92. Can we close this, at least for now?

mpenciak:
> I think the primary need this worked toward addressing is fulfilled by #92. Can we close this, at least for now?

Yeah, let's close this in favor of #92

PR #59, opened 2023-10-02
[refactor] use associated types instead of generics in traits where that makes sense

huitseeker:
Nova uses many traits, and more specifically many generic traits, even when those often only would have one single plausible implementation for a given struct.

This PR minimally modifies those traits that:
-  do clearly not aim to have several implementations per struct (`ROTrait`, `ROCircuitTrait`, `CommitmentEngineTrait`,`EvaluationEngineTrait`),
- involve generic constraints that introduce some complexity elsewhere in the implementation,

so as to drop the generic parameter, and replace it with an associated type.

Plenty of other traits which don't really aim at having several instances per struct are left unchanged (since arguably their only consequence in the Nova code base is needing a decoration with the generic parameter).

> [!NOTE]
> The following guide explains the mechanics and rationale of the change, note the paragraph on token structs:
> https://gist.github.com/huitseeker/6e5c92a2b17c96f29984223d53e59880

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -556,7 +556,7 @@ where
 
     // check if the output hashes in R1CS instances point to the right running instances
     let (hash_primary, hash_secondary) = {
-      let mut hasher = <<G2 as Group>::RO as ROTrait<G2::Base, G2::Scalar>>::new(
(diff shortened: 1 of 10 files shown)


huitseeker:
Closing momentarily in favor of other high priority tasks. Benefits at the moment are marginal

PR #70, opened 2023-10-13
refactor: Parametrization of lib.rs tests for various EvaluationEngines

huitseeker:
[microsoft/171](https://github.com/microsoft/Nova/pull/171) introduced enhanced genericity in tests to facilitate unit testing across diverse curve cycles.  Since then, the lib.rs tests, being designed to verify certain behaviors of the decider SNARK, tend to embed the use of a RelaxedR1CSSnark<G, EE>. They do this under the alias (where the test carefully chooses a specific import path for RelaxedR1CSSnark):

```rust
type S1<G> = RelaxedR1CSSnark<G, ipa_pc::EvaluationEgine<G>>
```
Effectively, this constrains the PCS utilized by that decider to the IPA, even if the only thing the test function needed to be specific about is `RelaxedR1CSSnark`.

Hence, these generic tests then had to accommodate constraints facilitating the use of the IPA, specifically:

```rust
<G::CE as CommitmentEngineTrait<G>>::CommitmentKey: CommitmentKeyExtTrait<G>
```
This is because the IPA operates exclusively with a "splittable" key, a concept delineated by the CommitmentKeyExtTrait.

However, two core adjustments are required:

- For the inclusion of different PCSs (for instance, Zeromorph), it's imperative to render `EE<G>: EvaluationEngineTrait<G>` configurable.
- Parametrizing the tests based on the `EvaluationEngineTrait` implementation eliminates boundaries associated with `CommitmentKeyExtTrait`. Such boundaries matching the `EvaluationEngine` and `CommitmentEngine` only materialize when the scheme is instantiated (i.e., during test invocation), a point where they are seamlessly met.

The present PR effects this parametrization, effectively making the lib.rs tests functions variable based on the evaluation engine in use.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -940,14 +940,14 @@ type CE<G> = <G as Group>::CE;
 #[cfg(test)]
 mod tests {
   use crate::provider::bn256_grumpkin::{bn256, grumpkin};
-  use crate::provider::pedersen::CommitmentKeyExtTrait;
   use crate::provider::secp_secq::{secp256k1, secq256k1};
+  use crate::traits::evaluation::EvaluationEngineTrait;
   use core::fmt::Write;
 
   use super::*;
   type EE<G> = provider::ipa_pc::EvaluationEngine<G>;
-  type S<G1> = spartan::snark::RelaxedR1CSSNARK<G1, EE<G1>>;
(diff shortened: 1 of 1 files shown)


mpenciak (approved):
Looks good!

Issue #35, opened 2023-08-31
ZeroMorph: Polish and document code


emmorais:
Add comment to make it easier to understand how the implements the paper. 

Consider renaming some variables.

huitseeker:
Closed in #142

Issue #22, opened 2023-08-26
ZeroMorph: Initialize Fiat-Shamir in a secure way


emmorais:
We don't want to be vulnerable to the Frozen Heart vulnerability: https://blog.trailofbits.com/2022/04/15/the-frozen-heart-vulnerability-in-bulletproofs/

More information here: https://eprint.iacr.org/2023/691

huitseeker:
Closed in #145

PR #71, opened 2023-10-13
Zeromorph EvaluationEngine trait

### LINKED ISSUE #19: Adjust ZeroMorph for EvaluationEngineTrait
KZG is a building block that allows us to commit to univariate polynomials. ZeroMorph uses that building block to construct a commitment to multilinear polynomials, which can be used by Nova. 

Part of the interfaces are already prepared for this trait, since polynomials and input points are represented as &[Fr], and the evaluation of this polynomial at this point is &Fr.

However, other adjustments are necessary. Namely, take a look on things like CommitmentEngineTrait, CommitmentKey, ProverKey, VerifierKey, etc.

huitseeker:
- makes tests generic in the `EvaluationEngine` by adopting #70,
- documents the gap between ZM / IPA: "endianness" w/ which coefficients must be supplied to the scheme,
- implements a `CommitmentKey` with a token universal setup, which allows implementing a `CommitmentEngineTrait` for KZG commitments,
- implements the prove & verify methods for the `EvaluationEngineTrait` implementation for ZMPCS. The scheme is slow (due to poor univariate KZG batching) 

Closes #19 

TODO: 
- [x] modify `gen_srs_for_testing` so the tests run in reasonable time.
- [x] cleaning up "point endianness" a bit

Next steps:
- making univariate KZG actually fast,
- document the protocol a bit more #35, 
- #22,

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1330,12 +1330,6 @@ mod tests {
       S<grumpkin::Point, EE<_>>,
     >();
     test_ivc_nontrivial_with_compression_with::<
-      bn256::Point,
-      grumpkin::Point,
-      S<bn256::Point, ZM<halo2curves::bn256::Bn256>>,
-      S<grumpkin::Point, EE<_>>,
-    >();
-    test_ivc_nontrivial_with_compression_with::<
       secp256k1::Point,
       secq256k1::Point,
       S<secp256k1::Point, EE<_>>,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1453,15 +1453,20 @@ mod tests {
 
     test_ivc_nontrivial_with_spark_compression_with::<G1, G2, EE<_>, EE<_>>();
     test_ivc_nontrivial_with_spark_compression_with::<
-      bn256::Point,
-      grumpkin::Point,
-      ZM<halo2curves::bn256::Bn256>,
-      EE<_>,
-    >();
-    test_ivc_nontrivial_with_spark_compression_with::<
       secp256k1::Point,
       secq256k1::Point,
       EE<_>,
+      EE<_>,
+    >();
+  }
+
+  #[test]
+  #[ignore]
+  fn test_ivc_nontrivial_with_spark_zm_compression() {
+    test_ivc_nontrivial_with_spark_compression_with::<
+      bn256::Point,
+      grumpkin::Point,
+      ZM<halo2curves::bn256::Bn256>,
       EE<_>,
     >();
   }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -948,13 +948,10 @@ mod tests {
   use core::fmt::Write;
 
   use super::*;
-  #[allow(dead_code)]
-  type ZM<E> = provider::non_hiding_zeromorph::ZMEvaluation<E>;
+  type ZM<E> = provider::non_hiding_zeromorph::ZMPCS<E>;
   type EE<G> = provider::ipa_pc::EvaluationEngine<G>;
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -1330,7 +1326,13 @@ mod tests {
     test_ivc_nontrivial_with_compression_with::<
       bn256::Point,
       grumpkin::Point,
-      S<bn256::Point, EE<_>>, // SZM<bn256::Point, halo2curves::bn256::Bn256>,
+      S<bn256::Point, ZM<halo2curves::bn256::Bn256>>,
+      S<grumpkin::Point, EE<_>>,
+    >();
+    test_ivc_nontrivial_with_compression_with::<
```
storojs72, line 1332: Why do we need this test instance to run twice using exactly same parameters? Shouldn't it better to use:
```
test_ivc_nontrivial_with_compression_with::<
      bn256::Point,
      grumpkin::Point,
      SPrime<bn256::Point, ZM<halo2curves::bn256::Bn256>>,
      SPrime<grumpkin::Point, EE<_>>,
    >();
```
instead, in order to use pp-spartan's proving / verifications for one of test instances?

storojs72 (commented):
Shouldn't we use pp-spartan by default in Arecibo, since we consider it more efficient and Ethereum-friendly than "regular" Spartan? At least I would propose mandatory having e2e proof/verify test instances in `lib.rs` for Bn256/Grumpkin curve cycle as we consider it as default in our solidity-verifier.

Trying to run `test_ivc_nontrivial_with_compression` with zeromorph stuff using pp-spartan, e.g.:
```
test_ivc_nontrivial_with_compression_with::<
      bn256::Point,
      grumpkin::Point,
      SPrime<bn256::Point, ZM<halo2curves::bn256::Bn256>>,
      SPrime<grumpkin::Point, EE<_>>,
    >();
```

gives me following assertion error (at setup phase):

```
thread 'tests::test_ivc_nontrivial_with_compression' panicked at 'assertion failed: ck.length() >= Self::commitment_key_floor()(S)', src/spartan/ppsnark.rs:912:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test tests::test_ivc_nontrivial_with_compression ... FAILED
```

src/lib.rs
huitseeker: Yep, that's silly!

huitseeker:
@storojs72 your point about the default doesn't really apply to Arecibo, since there isn't any "default" per se, Arecibo/Nova just provides the two APIs (compression, spark_compression) side-by-side. But it does apply to Lurk!

> Trying to run test_ivc_nontrivial_with_compression with zeromorph stuff using pp-spartan, e.g.:

Yes, that's normal and due to microsoft/nova#203, or #5 here. `test_ivc_nontrivial_with_compression` is meant to run with spartan but not computational commitment. The test you're looking for is `test_ivc_nontrivial_with_spark_compression` a few lines downstream. The reason it doesn't exhibit the same error is in the initialization of the public parameters.

src/lib.rs
```diff
@@ -1330,7 +1326,7 @@ mod tests {
     test_ivc_nontrivial_with_compression_with::<
       bn256::Point,
       grumpkin::Point,
-      S<bn256::Point, EE<_>>, // SZM<bn256::Point, halo2curves::bn256::Bn256>,
+      S<bn256::Point, ZM<halo2curves::bn256::Bn256>>,
```
huitseeker, line 1329: This translates `test_ivc_nontrivial_with_zm_compression`

mpenciak (approved):
Looks great!

Issue #67, opened 2023-10-11
Universalize CommitmentKey in Arecibo to save space and streamline implementation


huitseeker:
The nature of public params in Arecibo presents a key opportunity for improvement, particularly in relation to how many bases are used in the `CommitmentKey`, which bases are the same for any proof using the IPA PCS (i.e. before Zeromorph, all of them).

## Current Behavior:

The CommitmentKey acts as a universal parameter and ideally, we'd have a single large disk copy stored for big circuits, which can then be trimmed down as needed. The key difference between two Nova proofs (and possibly, the kind of SNARK they use) should be solely based on the number of elements they require.

## Desired Behavior:

Ensure a single, universal `CommitmentKey` is stored and utilized to trim down specifics for each proof.

## Advantages:

- This would be a crucial step towards implementing Zeromorph ([arecibo/issues/19](https://github.com/lurk-lab/arecibo/issues/19)). In the PCS system of Zeromorph, which mandates a universal setup, there's a clear demarcation between Public Parameters (universal) and the Prover Key (specific to proof).

- There's also an opportunity to save substantial space. In Nova, unlike SuperNova which builds the CommitmentKey separately, the CommitmentKey is encapsulated within public parameters. As a result, any serialization duplicates it. Mutualizing it among proofs will yield significant space savings ([lurk-rs/issues/387](https://github.com/lurk-lab/lurk-rs/issues/387)).

winston-h-zhang:
I think there's a growing argument here to move the Lurk config/cache infra to `arecibo`. The main reason why is because it would make the most sense for `arecibo` to be responsible for its own commitment keys. For example, having the infra in Lurk wouldn't help if you needed to run many tests that needs to regenerate keys in `arecibo`. A secondary reason is to manage the memory pressure of public params. We could have the commitment key separate and only allocate it in when we need to commit something. What do you think?

PR #79, opened 2023-10-22
A cache for saving public params and commitment keys to local disk

winston-h-zhang:
Addresses #67.

This PR is largely a port of the already existing infrastructure from `lurk-rs`. We implement `Cache` as an interface to read, write, and update public parameters and commitment keys as the user needs. We also declare a global `ARECIBO_CACHE` for any downstream `arecibo` users to access, although users can also just declare a cache themselves. 

- We modify the `Group` trait by adding a way to extend commitment keys without having to recompute the entire string. 
- We add a new `CommitmentKeyTrait` to define the behavior of commitment keys.


src/cache.rs
```diff
@@ -0,0 +1,300 @@
+//! This module provides caching utils
```
huitseeker, line 1: Here's some inspiration to solve your imminent issue of compiling on WASM in a transparent manner:
https://github.com/rust-lang/rust/blob/master/compiler/rustc_data_structures/src/memmap.rs

src/traits/circuit_supernova.rs
```diff
@@ -25,6 +27,15 @@ pub trait StepCircuit<F: PrimeField>: Send + Sync + Clone {
     pc: Option<&AllocatedNum<F>>,
     z: &[AllocatedNum<F>],
   ) -> Result<(Option<AllocatedNum<F>>, Vec<AllocatedNum<F>>), SynthesisError>;
+
+  /// AHHH
```
huitseeker, line 31: This feels like it's halfway towards a confirmation of [Strehmel's results](https://cme.h-its.org/exelixis/pubs/JanThesis.pdf). Pray tell us more.

src/traits/mod.rs
```diff
@@ -80,7 +80,17 @@ pub trait Group:
   fn preprocessed(&self) -> Self::PreprocessedGroupElement;
 
   /// Produce a vector of group elements using a static label
-  fn from_label(label: &'static [u8], n: usize) -> Vec<Self::PreprocessedGroupElement>;
+  fn from_label(label: &'static [u8], n: usize) -> Vec<Self::PreprocessedGroupElement> {
+    Self::from_label_with_offset(label, 0, n)
+  }
+
+  /// Produce a vector of `n - offset` group elements using a static label,
+  /// by first generating `n` elements and then discarding the first `offset` of them
```
huitseeker, line 88: This feels like it makes an assumption those elements are *deterministically* generated from the label, and this should be documented very explicitly for the offset to make sense.

src/constants.rs
```diff
@@ -8,3 +8,9 @@ pub(crate) const NUM_FE_FOR_RO: usize = 24;
 
 /// Bit size of Nova field element hashes
 pub const NUM_HASH_BITS: usize = 250;
+
+/// The threshold of elements we check when extending a `CommitmentKey`
+/// to be convinced that they were generated from the same label
+///
+/// TODO: I have no idea about *any* security implications here.
```
huitseeker, line 15: In a caching context, you should either :
- trust your local disc, in which case the checking threshold should be 0, or 1 to avoid regressions due to code changes, or
- not trust the cache, in which case the appropriate threshold is INFINITY

src/cache.rs
```diff
@@ -0,0 +1,305 @@
...
+
+impl Cache {
+  /// Sets the inner root directory of the cache
+  pub fn set_inner(&self, path: &PathBuf) {
+    self.inner.lock().unwrap().clone_from(path);
+
+    create_dir_all(path).unwrap();
+    create_dir_all(path.join("commitment_keys")).unwrap();
+    create_dir_all(path.join("public_params")).unwrap();
+  }
```
huitseeker, line 69: There's no way to get a cache other than `init_cache()`, which makes me think this should be called `reset_inner`. But as mentioned above, I don't think we need the functionality at this stage.

src/cache.rs
```diff
@@ -0,0 +1,305 @@
...
+
+  /// Returns a commitment key from the cache. If the cached commitment key is shorter
+  /// than what we need, we extend the key and update the cache as a side effect.
+  ///
+  /// This is the same as calling `G::CE::setup(b"ck", n)`
+  pub fn commitment_key<G: Group>(&self, n: usize) -> io::Result<CommitmentKey<G>> {
+    let mut file = self.ck_file::<G>()?;
+
+    let ck_size = file.read_u64::<LittleEndian>()? as usize;
+    println!("ck_size: {}, n: {}", ck_size, n);
```
huitseeker, line 123: Please use the `log` crate instead.

src/cache.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    let mut mmap = unsafe { MmapMut::map_mut(&file)? };
+    let skip = std::mem::size_of::<usize>();
+    let bytes = &mut mmap[skip..]; // skip the first `usize` bytes
```
huitseeker, line 127: Ah, so we should at least make clear that despite the use of concurrency-safe structures above, this is *not* concurrency-safe, right? Two calls to `commitment_key` (including those generated as simply as running unit tests concurrently) could easily step on each other's toes.

Have you considered something like a `struct MemoryMappedMutFile(Arc<RwLock<MmapMut>>);` of which the cache would create a singleton instance per relevant file (commitment key, public params)?

src/cache.rs
```diff
@@ -0,0 +1,305 @@
...
+mod tests {
+  use std::path::PathBuf;
+
+  use pasta_curves::pallas;
+
+  use crate::{
+    provider::{bn256_grumpkin::bn256, secp_secq::secq256k1},
+    traits::{commitment::CommitmentEngineTrait, Group},
+  };
+
```
huitseeker, line 280: I think the concurrency test is three threads all trying to extend the same CK, here.

src/provider/pedersen.rs
```diff
@@ -33,10 +34,40 @@ impl<G: Group> Clone for CommitmentKey<G> {
...
+  }
+
+  fn encode<W: std::io::Write>(&self, write: &mut W) -> io::Result<()> {
+    // TODO: unwrap()
+    unsafe { abomonation::encode(self, write) }
+  }
+
+  fn decode(bytes: &mut [u8], n: usize) -> io::Result<Self> {
+    // Note: this is unsafe but stable; it only assumes that `bytes` is any valid commitment key
+    unsafe {
```
huitseeker, line 53: I think this unsafe section can be smaller.

src/provider/pedersen.rs
```diff
@@ -33,10 +34,40 @@ impl<G: Group> Clone for CommitmentKey<G> {
...
+      // Ideally, we would like to allocate only one commitment key, because if we simply loaded
+      // all the bytes from the reader, we might be tapping into a cached commitment key with
+      // >=10^7 elements and kill the process. Instead, we need to manipulate the bytes to convince
+      // `abomonation::decode` to only decode what we need by manipulating the `Vec` header.
+      let mut test_ck = Self { ck: Vec::new() };
+      test_ck.ck.set_len(n);
+      let header_size = mem::size_of::<Self>();
+      let header: &[u8] =
+        std::slice::from_raw_parts(&test_ck as *const _ as *const u8, header_size);
+      bytes[..header_size].copy_from_slice(header);
```
huitseeker, line 63: I suspect this can fit within a `MaybeUninit` usage: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-an-array-element-by-element

src/traits/commitment.rs
```diff
@@ -101,6 +115,33 @@ pub trait CommitmentEngineTrait<G: Group>: Clone + Send + Sync {
   /// Samples a new commitment key of a specified size
   fn setup(label: &'static [u8], n: usize) -> Self::CommitmentKey;
 
+  /// Internal implementation of [CommitmentEngineTrait::extend]
+  ///
+  /// **Do not call this directly**. This does not check the consistency of the given label.
```
huitseeker, line 120: One way to improve this is to call us `extend_aux_unsafe`

huitseeker (commented):
I think there's a concurrency issue in the access of the CK / PPs. Happy to set some time to discuss it along with potential solutions.

huitseeker (changes requested):
For emphasis, this is the important comment:
https://github.com/lurk-lab/arecibo/pull/79#discussion_r1371838518

winston-h-zhang:
Closed in favor of #335, this has bit-rot too much

PR #80, opened 2023-10-23
Supernova Compressed SNARKs

mpenciak:
This PR implements a final compressing SNARK for Supernova NIVC proofs. It also adds a test for the compressed SNARK.

This version 0 of compression simply generates a SNARK proof for each primary running instance, and the secondary instance. The verifier then checks the validity of each of the SNARKs.

### DIFF
--- a/src/supernova/snark.rs
+++ b/src/supernova/snark.rs
@@ -301,24 +301,259 @@ fn field_as_usize<F: PrimeField>(x: F) -> usize {
 #[cfg(test)]
 mod test {
   use super::*;
-  use crate::provider::{
-    bn256_grumpkin::{bn256, grumpkin},
-    secp_secq::{secp256k1, secq256k1},
+  use crate::{
+    provider::{
+      bn256_grumpkin::{bn256, grumpkin},
+      ipa_pc::EvaluationEngine,
+      secp_secq::{secp256k1, secq256k1},
+    },
+    spartan::snark::RelaxedR1CSSNARK,
+    supernova::NonUniformCircuit,
+    traits::{circuit_supernova::TrivialSecondaryCircuit, evaluation::EvaluationEngineTrait},
   };
+
+  use abomonation::Abomonation;
+  use bellpepper_core::{num::AllocatedNum, ConstraintSystem, SynthesisError};
+  use ff::Field;
   use pasta_curves::{pallas, vesta};
 
-  fn test_nivc_nontrivial_with_compression_with<G1, G2>()
+  type EE<G> = EvaluationEngine<G>;
+  type S<G, EE> = RelaxedR1CSSNARK<G, EE>;
+
+  #[derive(Clone)]
+  struct SquareCircuit<G: Group> {
+    _p: PhantomData<G>,
+  }
+
+  impl<G: Group> StepCircuit<G::Scalar> for SquareCircuit<G> {
+    fn arity(&self) -> usize {
+      1
+    }
+
+    fn circuit_index(&self) -> usize {
+      0
+    }
+
+    fn synthesize<CS: ConstraintSystem<G::Scalar>>(
+      &self,
+      cs: &mut CS,
+      _pc: Option<&AllocatedNum<G::Scalar>>,
+      z: &[AllocatedNum<G::Scalar>],
+    ) -> Result<
+      (
+        Option<AllocatedNum<G::Scalar>>,
+        Vec<AllocatedNum<G::Scalar>>,
+      ),
+      SynthesisError,
+    > {
+      let z_i = &z[0];
+
+      let z_next = z_i.square(cs.namespace(|| "z_i^2"))?;
+
+      let next_pc = AllocatedNum::alloc(cs.namespace(|| "next_pc"), || Ok(G::Scalar::from(1u64)))?;
+
+      cs.enforce(
+        || "next_pc = 1",
+        |lc| lc + CS::one(),
+        |lc| lc + next_pc.get_variable(),
+        |lc| lc + CS::one(),
+      );
+
+      Ok((Some(next_pc), vec![z_next]))
+    }
+  }
+
+  #[derive(Clone)]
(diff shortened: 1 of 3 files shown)


src/supernova/snark.rs
```diff
@@ -0,0 +1,559 @@
...
+      &recursive_snark.l_w_secondary,
+    );
+
+    let (nifs_secondary, (f_U_secondary, f_W_secondary)) = res_secondary?;
+
+    let r_W_snark_primary = pk
+      .pks_primary
+      .iter()
+      .zip(&recursive_snark.r_W_primary)
+      .zip(&recursive_snark.r_U_primary)
```
porcuquine, line 144: I believe that this will produce a result that is shorter than `pk.pks_primary` if `recursive_snark.r_W_primary` or `recursive_snark.r_U_primary` are. That *could* be okay, since it will end up being an invalid `CompressedSNARK` as the result of trying to compress an invalid `RecursiveSNARK`. However, it's probably better to catch the problem and refuse to produce an invalid proof.

Note: it's extremely important that the *verifier* refuse to verify such invalid `CompressedSNARK`s — which I believe this initial implementation fails to do. I will comment below in more detail so you can verify.

src/supernova/snark.rs
```diff
@@ -0,0 +1,559 @@
...
+    let res_primary = self
+      .r_W_snark_primary
+      .iter()
+      .enumerate()
+      .zip(&vk.vks_primary)
+      .map(|((idx, proof), vk)| {
+        let U = &self.r_U_primary[idx];
+        proof.verify(vk, U)
+      })
+      .collect::<Vec<_>>();
```
porcuquine, line 272: This only produces as many results as there are values in `self.r_W_snark_primary`, even if `vk.vks_primary` has more.

src/supernova/snark.rs
```diff
@@ -0,0 +1,559 @@
...
+    )?;
+
+    let res_secondary = self
+      .f_W_snark_secondary
+      .verify(&vk.vk_secondary, &f_U_secondary);
+
+    res_primary
+      .iter()
+      .map(|res| res.clone().map_err(|_| NovaError::ProofVerifyError))
+      .collect::<Result<Vec<_>, _>>()?;
```
porcuquine, line 288: This will error if a bad result has been stored in `res_primary`, but it won't catch the case in which some checks were never performed. Test case to consider: is it possible to provide input to this verification function such that `res_primary` is empty here? If so, would that input represent a valid proof? If not, would this verification function successfully 'fail' (by returning an error)?

It would be a good exercise and provide valuable confidence in verifier soundness to write one or more 'negative' tests, demonstrating that such maliciously-constructed `CompressedSNARK`s will not wrongly verify.

porcuquine (changes requested):
This is a partial review, noting a potential soundness issue. This is proximately due to a potential subtle bug relating to the behavior of `zip` (truncating results to the length of the shorter of its inputs). It's more broadly due to a failure to explicitly check a condition of the `CompressedSNARK` that seems critical: specifically, we need to not only check that all provided 'sub-SNARKs' verify, but also that all *required* sub-SNARKs were provided.

Whether or not the problem I've pointed to is manifested here, let's add a test conclusively demonstrating that the next iteration does *not* have that issue.

src/supernova/snark.rs
```diff
@@ -207,6 +215,8 @@ where
     + (7 + 2 * pp.augmented_circuit_params_primary.get_n_limbs()); // # 1 * (7 + [X0, X1]*#num_limb)
 
     // secondary circuit
+    // NOTE: This count ensure the number of witnesses sent by the prover must bd equal to the number
+    // of NIVC circuits
     let num_field_secondary_ro = 2 // params_next, i_new
     + 2 * pp.circuit_shape_secondary.F_arity // zo, z1
     + pp.circuit_shapes.len() * (7 + 2 * pp.augmented_circuit_params_primary.get_n_limbs()); // #num_augment
```
mpenciak, line 222: > This will error if a bad result has been stored in `res_primary`, but it won't catch the case in which some checks were never performed. Test case to consider: is it possible to provide input to this verification function such that `res_primary` is empty here? If so, would that input represent a valid proof? If not, would this verification function successfully 'fail' (by returning an error)?
> 
> It would be a good exercise and provide valuable confidence in verifier soundness to write one or more 'negative' tests, demonstrating that such maliciously-constructed `CompressedSNARK`s will not wrongly verify.
https://github.com/lurk-lab/arecibo/pull/80#discussion_r1373790940

(not intentionally, but...) This count for the RO initialization actually detects a proof that is invalid because the verifier is not given the right number of running instances. 

I have added a corresponding test to ensure that this is indeed happening.

porcuquine: Ah, great. Thank you for clarifying that detail.

porcuquine: My points above stand — in terms of UX and subtle probably-accidental implementation pitfalls; but I acknowledge that this check may make my point about verifier correctness inapplicable. Thanks for adding a test to demonstrate this.

src/supernova/snark.rs
```diff
@@ -0,0 +1,656 @@
...
+        _ => panic!("Invalid circuit index"),
+      }
+    }
+
+    fn secondary_circuit(&self) -> TrivialSecondaryCircuit<G2::Scalar> {
+      Default::default()
+    }
+  }
+
+  fn test_nivc_trivial_with_compression_with<G1, G2, E1, E2>()
```
porcuquine, line 496: We should make sure to add `CompressedSNARK` tests when more NIVC tests become available (#52).

porcuquine:
Since we're moving forward with the v1 `CompressedSNARK` before upstreaming SuperNova, this PR can be closed without merging. @mpenciak feel free to leave open until then, or close sooner if you prefer.

mpenciak:
Now that PR #108 is open, I think it's safe to close this PR.

Issue #53, opened 2023-09-25
Manage `CommitmentKeyHint`s properly in supernova


winston-h-zhang:
This is where the optfn1, optfn2 parameters from above come in. They need to be parametrized by a preference expressed as a static function of the SNARK that will eventually be used to compress the proof. See #5. We should open an issue to fix this across SuperNova.

_Originally posted by @huitseeker in https://github.com/lurk-lab/arecibo/pull/51#discussion_r1334773944_

huitseeker:
Closed in #95

PR #81, opened 2023-10-25
Merge Nova and SuperNova types

### LINKED ISSUE #69: Unify `nova::PublicParams` and `supernova::PublicParams`
`supernova::PublicParams` will become the only `PublicParams` of `nova_snark`. The current `nova::PublicParams` will be removed, and instead be replaced by the special case of `supernova::PublicParams` where `pp.circuit_shapes.len() == 1`.

winston-h-zhang:
Resolves #69 and #53.

This PR merges some of the common types between Nova and SuperNova. Notably, `PublicParams` is merged. We take the base SuperNova type to be the new common use, since they generalize the corresponding Nova type. We also merge the `{Nova,SuperNova}AugmentedCircuitParams` -- they were identical.

Other cleanups:
- We collect and organize all the public parameter infrastructure that has been developed into `src/parameters.rs`.
- We add `<C1, C2>` bounds to `supernova::RecursiveSNARK` to remove the weird inconsistency between `nova::RecursiveSNARK`.
- We add `CommitmentKeyHint` when creating SuperNova public parameters.


src/traits/circuit.rs
```diff
@@ -11,22 +11,87 @@ pub trait StepCircuit<F: PrimeField>: Send + Sync + Clone {
...
+/// A trivial step circuit that simply returns the input, for use on the secondary circuit when implementing NIVC.
+/// NOTE: This should not be needed. The secondary circuit doesn't need the program counter at all.
+/// Ideally, the need this fills could be met by `traits::circuit::TrivialTestCircuit` (or equivalent).
+#[derive(Clone, Debug, Default)]
+pub struct TrivialSecondaryCircuit<F: PrimeField> {
   _p: PhantomData<F>,
 }
 
-impl<F> StepCircuit<F> for TrivialCircuit<F>
+impl<F> StepCircuit<F> for TrivialSecondaryCircuit<F>
```
porcuquine, line 94: I think we should prefer`TrivialCircuit` when combining these.

benches/compute-digest.rs
```diff
@@ -54,7 +54,7 @@ where
     }
   }
 }
-impl<F> StepCircuit<F> for NonTrivialCircuit<F>
+impl<F> StepCircuit<F> for NonTrivialTestCircuit<F>
```
huitseeker, line 57: FYI, this is just a renaming, but https://github.com/microsoft/Nova/commit/fb83e30e16e56b3b21c519b15b83c4ce1f077a13 makes it unlikely it will be accepted upstream. We should normalize to `NonTrivialCircuit`.

src/traits/snark.rs
```diff
@@ -23,9 +25,9 @@ pub trait RelaxedR1CSSNARKTrait<G: Group>:
   /// a minimum sizing cue for the commitment key used by this SNARK
   /// implementation. The commitment key passed in setup should then
   /// be at least as large as this hint.
-  fn commitment_key_floor() -> Box<dyn for<'a> Fn(&'a R1CSShape<G>) -> usize> {
+  fn commitment_key_floor() -> Arc<dyn for<'a> Fn(&'a R1CSShape<G>) -> usize> {
```
huitseeker, line 28: Why the Arc? This is a stateless function.

src/parameters.rs
```diff
@@ -0,0 +1,508 @@
...
+
+  /// Compute primary and secondary commitment keys sized to handle the largest of the circuits in the provided
+  /// `CircuitShape`.
+  fn compute_primary_ck(
+    circuit_params: &[CircuitShape<G1>],
+    optfn1: Option<CommitmentKeyHint<G1>>,
+  ) -> CommitmentKey<G1> {
+    let size_primary = circuit_params
+      .iter()
+      .map(|circuit| commitment_key_size(&circuit.r1cs_shape, optfn1.clone()))
```
winston-h-zhang, line 453: `Arc` because of this `.clone()` here @huitseeker

porcuquine (dismissed):

porcuquine (approved):

PR #86, opened 2023-10-28
Port paranova.

porcuquine:
This is a port of the much-appreciated ParaNova prototype from
https://github.com/privacy-scaling-explorations/Nova/pull/11

You can learn more about Paranova from https://zkresear.ch/t/parallelizing-nova-visualizations-and-mental-models-behind-paranova/198 or https://zuzalu.streameth.org/session/169

The idea is to use the general shape and scaffolding to iterate on in arecibo. Since those plans are beyond the scope of this port, I won't elaborate here.

For this to be accepted, we should primarily feel confident that it doesn't interfere with or harm existing Nova/SuperNova work. Over time, either before or after merging this, we can work on refactoring so Nova is least affected and (to the extent it must be) intelligently modified to best support the work here while staying true to its nature.

Note that Nova and SuperNova tests still pass and circuit constraint count has not increased. This work intends not to affect the substance of existing circuits or proving mechanisms.

h/t @oskarth @chiro-hiro @aleph-v @nalinbhardwaj @CPerezz


aleph-v:
Hey, umm so I have 0 context on this but I was tagged here. If you're importing the code from the prototypes we worked in the Vietnam 0xParc residency I would advise some caution as I wrote a large portion of it and I know for a fact it has soundness bugs.

porcuquine:
Thanks for your diligence. We're going to include the important incremental work by @nikkolasg (https://github.com/nikkolasg/Nova/blob/mapreduce/examples/merkle_tree.rs) as described here: https://lagrangelabs.notion.site/Nova-from-IVC-to-general-PCD-for-zkMapReduce-d04afec2d70c47db9dfb67e84d0d796e. This was an exercise in bringing your prototype up-to-date, and we are aware that it has issues.

Lev-Stambler:
I would love to help this get up and running!

Issue #27, opened 2023-08-26
Compressed SNARK for SuperNova


porcuquine:
#2 introduces SuperNova but does not provide a compressed SNARK implementation.

v0 could require a separate compressed SNARK for each circuit's final witness — with size and verification costs linear in the number of sub-circuits.

v1 should provide a single SNARK that proves all sub-circuits as one larger circuit.

PR #92, opened 2023-11-02
Add non-deterministic ivc test.

porcuquine:
This PR adds a test exercising support for non-deterministic hints in NIVC. It is based on the equivalent test for IVC in Nova. Instead of proving a series of sequential fifth-root computations (which would be expensive to calculate directly, so instead are calculated in reverse to produce the sequence of hints) — we produce a sequence of alternating cube and fifth roots. This alternation exercises NIVC. We hardcode the next program-counter into the circuit, so that alternation is enforce.

The other changes are fixing typos in the code ('unifrom' -> 'uniform') and removing the default implementation of `circuit_index` for `circuit_supernova::StepCircuit` — which proved error-prone (confirming a suggestion @arthurpaulino made previously).

TODO:
- Add a test of compression once the NIVC CompressedSNARK lands. (#27 #80)

### DIFF
--- a/benches/recursive-snark-supernova.rs
+++ b/benches/recursive-snark-supernova.rs
@@ -339,6 +339,10 @@ where
     1
   }
 
+  fn circuit_index(&self) -> usize {
+    0
+  }
+
   fn synthesize<CS: ConstraintSystem<F>>(
     &self,
     cs: &mut CS,
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -168,8 +168,8 @@ where
   C2: StepCircuit<G2::Scalar>,
 {
   /// Construct a new [PublicParams]
-  pub fn new<NC: NonUniformCircuit<G1, G2, C1, C2>>(non_unifrom_circuit: &NC) -> Self {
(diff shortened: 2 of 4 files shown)


mpenciak (approved):
Merging the `NonUniformCircuit` and `StepCircuit` data types was something I was missing in the examples in #52. This strategy of writing the main loop over the circuits themselves also seems like the better way of doing things.

In #52 I was using a combination of getting the program counter via `recursive_snark.get_program_counter()` and then turning it into an `augmented_circuit_index` and finally calling `prove_step` using `nonuniform_circuit.primary_circuit(augmented_circuit_index)`. This avoids all of the juggling of program counters.

I've convinced myself that these two approaches are equivalent, but I can see a (somewhat contrived) situation where the second approach might still come up:

If the circuit writer is exceptionally lazy, and just wants to calculate the "`next_pc`" in the circuit, and then write their proving loop by choosing which circuit synthesize at each step using that calculated program counter. 

Anyway this is just my personal musings about the different shapes that the Supernova API allows the user to write their NIVC calculations. I'm approving (pending the final addition of a `CompressedSNARK` part of the test)

PR #95, opened 2023-11-03
refactor: Refactor public parameters setup to remove optionality of commitment key size hint

huitseeker:
## Context

See #5 and https://github.com/microsoft/Nova/pull/203, by which public parameter creation takes a size hint. This size hint carries the particular requirements of the final compressing SNARK the user expected to use with these public parameters.

This avoids setting the size of public parameters too conservatively for all users of the code base, but rather limits the use of "big" public parameters to those constructions that will eventually need them.

## In this PR

We slightly simplify the API of passing a commitment key size hint when creating public parameters. We apply this to Supernova, closing #53.

Companion Lurk PR at https://github.com/lurk-lab/lurk-rs/pull/843

## In detail

- Converted `CommitmentKeyHint<G>` from a boxed dynamic trait object to a direct dynamic trait object in `r1cs/mod.rs`.
- Changed the `commitment_key` function to always require a commitment key floor, eliminating the need for default behavior when a floor function isn't provided.
- Updated the `r1cs_shape` function across various files to take in a `CommitmentKeyHint` instead of it being optional and introduce a closure as an argument.
- Relevant modifications and updates were made in the `r1cs_shape` and `commitment_key` function calls within the test functions for various modules.
- Ported use of commitment key hint to Supernova, closing #53.
- This PR puts Arecibo in line with the latest state of https://github.com/microsoft/Nova/pull/203

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,18 +140,21 @@ where
   ///
   /// # Note
   ///
-  /// Some SNARKs, like variants of Spartan, use computation commitments that require
-  /// larger sizes for some parameters. These SNARKs provide a hint for these values by
+  /// Public parameters set up a number of bases for the homomorphic commitment scheme of Nova.
+  ///
+  /// Some final compressing SNARKs, like variants of Spartan, use computation commitments that require
+  /// larger sizes for these parameters. These SNARKs provide a hint for these values by
   /// implementing `RelaxedR1CSSNARKTrait::commitment_key_floor()`, which can be passed to this function.
-  /// If you're not using such a SNARK, pass `None` instead.
+  ///
+  /// If you're not using such a SNARK, pass `&(|_| 0)` instead.
   ///
   /// # Arguments
   ///
   /// * `c_primary`: The primary circuit of type `C1`.
   /// * `c_secondary`: The secondary circuit of type `C2`.
-  /// * `optfn1`: An optional `CommitmentKeyHint` for `G1`, which is a function that provides a hint
+  /// * `ck_hint1`: A `CommitmentKeyHint` for `G1`, which is a function that provides a hint
   ///   for the number of generators required in the commitment scheme for the primary circuit.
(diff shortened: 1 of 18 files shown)


src/supernova/mod.rs
```diff
@@ -171,6 +171,23 @@ where
   C2: StepCircuit<G2::Scalar>,
 {
   /// Construct a new [PublicParams]
+  ///
+  /// # Note
+  ///
+  /// Public parameters set up a number of bases for the homomorphic commitment scheme of Nova.
+  ///
+  /// Some final ocmpressing SNARKs, like variants of Spartan, use computation commitments that require
```
adr1anh, line 179: Typo: "compressing"

adr1anh (commented):
Changes look good to me. Only noted a typo in the comments.

Idea: Define a default `CommitmentKeyHint` which just returns 0, instead of having `&(|_| 0)`.

huitseeker:
> Idea: Define a default CommitmentKeyHint which just returns 0, instead of having &(|_| 0).

Done!

src/lib.rs
```diff
@@ -169,17 +172,17 @@ where
...
   ///
   /// let pp = PublicParams::new(&circuit1, &circuit2, pp_hint1, pp_hint2);
   /// ```
   pub fn new(
     c_primary: &C1,
     c_secondary: &C2,
-    optfn1: Option<CommitmentKeyHint<G1>>,
-    optfn2: Option<CommitmentKeyHint<G2>>,
+    optfn1: &CommitmentKeyHint<G1>,
+    optfn2: &CommitmentKeyHint<G2>,
```
mpenciak, line 185: Not important, but now that it's not optional maybe we can rename these to `ck_hint1` and `ck_hint2` to match Supernova?

mpenciak (approved):
Looks good other than a small unimportant naming thing!

PR #97, opened 2023-11-03
Ensure exactly one selection in get_from_vec_alloc_relaxed_r1cs.

### LINKED ISSUE #94: Circuit soundness bug: unchecked program counter
The function [get_from_vec_alloc_relaxed_r1cs](https://github.com/lurk-lab/arecibo/blob/f81bd4c30a36811b36c5ab93bba875569f4df1b9/src/supernova/utils.rs#L15) does not check that `program_counter` is valid (i.e. in the range { $0,1,\ldots, \ell-1 $}. If `program_counter` is outside this range, the function returns the first `RelaxedR1CSInstance` by default. 

This can be fixed by enforcing that the sum of all [`equal_bits`](https://github.com/lurk-lab/arecibo/blob/f81bd4c30a36811b36c5ab93bba875569f4df1b9/src/supernova/utils.rs#L33) (including one for the first instance) is 1. This ensures that exactly one of the `equal_bits` is set, and therefore `program_counter` is in the correct range. 
Another possibility would be to ensure the returned `index` is equal to the input `target_index`.

Since the returned `index` is the same as the input `target_index`, it is not necessary for the function to return it. Instead, the function could return a vector of the `equal_bits`, which are already recomputed during the computation of [U_next](https://github.com/lurk-lab/arecibo/blob/f81bd4c30a36811b36c5ab93bba875569f4df1b9/src/supernova/circuit.rs#L397-L398).

porcuquine:
This PR fixes #94. I think the alternate approach (checking consistency of the chosen and received target index) is probably a little better. I may take a shot at this along with investigating a refactoring to avoid duplication of the bit-checking, etc. in a separate PR. Meanwhile, this should fix the immediate issue straightforwardly.

### DIFF
--- a/src/gadgets/utils.rs
+++ b/src/gadgets/utils.rs
@@ -484,3 +484,30 @@ pub fn select_num_or_one<F: PrimeField, CS: ConstraintSystem<F>>(
 
   Ok(c)
 }
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+  use crate::r1cs::util::FWrap;
+  use bellpepper_core::test_cs::TestConstraintSystem;
+  use pasta_curves::pallas::Scalar as Fr;
+  use proptest::prelude::*;
+
+  proptest! {
+    #[test]
+    fn test_enforce_alloc_num_equal_const((a, b) in any::<(FWrap<Fr>, FWrap<Fr>)>()) {
+      prop_assume!(a != b);
+
+        let test_a_b = |a, b| {
+            let mut cs = TestConstraintSystem::<Fr>::new();
+            let a_num = AllocatedNum::alloc_infallible(cs.namespace(|| "a_num"), || a);
+            let r = alloc_num_equals_const(&mut cs, &a_num, b);
+            assert_eq!(r.unwrap().get_value().unwrap(), a==b);
+        };
+        // negative testing
+        test_a_b(a.0, b.0);
+        // positive testing
+        test_a_b(a.0, a.0);
+    }
+  }
+}
--- a/src/gadgets/utils.rs
+++ b/src/gadgets/utils.rs
@@ -196,6 +196,48 @@ pub fn alloc_num_equals<F: PrimeField, CS: ConstraintSystem<F>>(
   Ok(r)
 }
 
+/// Check that two numbers are equal and return a bit
+pub fn alloc_num_equals_const<F: PrimeField, CS: ConstraintSystem<F>>(
+  mut cs: CS,
+  a: &AllocatedNum<F>,
+  b: F,
+) -> Result<AllocatedBit, SynthesisError> {
+  // Allocate and constrain `r`: result boolean bit.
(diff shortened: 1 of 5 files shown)


adr1anh:
I think with this we can remove [this comment](https://github.com/lurk-lab/arecibo/blob/f81bd4c30a36811b36c5ab93bba875569f4df1b9/src/supernova/circuit.rs#L420-L437). Checking the sum of the `equal_bits` is equivalent to range-check, so the comment no longer applies.

porcuquine:
Done.

src/supernova/test.rs
```diff
@@ -658,7 +658,7 @@ fn test_recursive_circuit() {
   let ro_consts1: ROConstantsCircuit<G2> = PoseidonConstantsCircuit::default();
   let ro_consts2: ROConstantsCircuit<G1> = PoseidonConstantsCircuit::default();
 
-  test_recursive_circuit_with::<G1, G2>(&params1, &params2, ro_consts1, ro_consts2, 9836, 12035);
+  test_recursive_circuit_with::<G1, G2>(&params1, &params2, ro_consts1, ro_consts2, 9835, 12028);
```
porcuquine, line 661: yay

porcuquine:
I played around, and this was the best I could do for constraints. I think it's a good stopping point. Now we reuse the selector, and I added an optimization for allocation of equality with constant.

huitseeker (commented):
Looks good to me! The comments on the multicase could benefit from being more descriptive about what they designate (this should be clear to Lurk developers, but I'm thinking about the nova context where we'd upstream this code).

src/gadgets/utils.rs
```diff
@@ -200,6 +200,48 @@ pub fn alloc_num_equals<F: PrimeField, CS: ConstraintSystem<F>>(
...
+
+  // Allocate t s.t. t=1 if z1 == z2 else 1/(z1 - z2)
+
+  let t = AllocatedNum::alloc(cs.namespace(|| "t"), || {
+    Ok(if *a.get_value().get()? == b {
+      F::ONE
+    } else {
+      (*a.get_value().get()? - b).invert().unwrap()
+    })
+  })?;
```
adr1anh, line 226: In situations where we want to compute a boolean selector vector, we may want to leverage `batch_invert` when computing the `t`s for each bit. Inversion is an expensive operation, and the threshold at which `batch_invert` becomes cheaper than individual inversions is very low (each additional element to the batch is only 3 field multiplications.

adr1anh (dismissed):
The second optimization round looks good to me. Left a small suggestion for the future "case" gadget.

src/gadgets/utils.rs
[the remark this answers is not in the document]
porcuquine: The latter doesn't quite work because of `subtle` stuff, fwiw.

huitseeker (approved):

Issue #64, opened 2023-10-09
`supernova::RecursiveSNARK::prove_step()` shouldn't need `augmented_circuit_index` as a parameter


mpenciak:
The current setup with the program counter for `supernova::RecursiveSNARK` makes it so that at each step of the proof the user has to provide an `augemented_circuit_index` to `RecursiveSNARK::prove_step()`.

`RecursiveSNARK` already internally contains the data of the program counter which has to be extracted from and then, immediately afterwards, provided as an input to `RecursiveSNARK` in the proving loop using some sort of logic similar to what is done [here](https://github.com/lurk-lab/arecibo/blob/cc71054d45855f0a027c3aaff4b9f1c072187b22/examples/nivc_nontrivial.rs#L254-L256).

If `RecursiveSNARK` internally kept track of the next augmented circuit index, it should be possible to change the signature of `prove_step` to:

```rust
  pub fn prove_step<N: NonUniformCircuit<G1, G2, C1, C2>>(
    &mut self,
    pp: &PublicParams<G1, G2, C1, C2>,
    non_uniform_circuit: &N,
    z0_primary: &[G1::Scalar],
    z0_secondary: &[G2::Scalar],
  ) -> Result<(), SuperNovaError>
```

and add the logic from the above-linked example to the `prove_step` method.

By making this change, we will probably be excluding some unusual use-cases for Supernova's NIVC, but I don't think it's excluding the main use-case where `program_counter` directly points to the next expected circuit to evaluate.

adr1anh:
On a similar note, the `iter_base_step` takes 3 different variations of "program counter". 

The function takes three counters:
- `circuit_index: usize`: used to index into the public parameters and get the primary circuit's shape. It is the program counter for the first curve. It is also then given to the second circuit as witness to indicate which accumulator to initialize. 
- `initial_program_counter: Option<G1::Scalar>`: Indicates which function is run in the first iteration on the primary circuit. It is only used at input to the first circuit, as should be the same as `circuit_index`
- `first_augmented_circuit_index: usize`: It says which one of the primary circuits should be folded. Again, same as `circuit_index`

Unless I'm missing something, having only a single `circuit_index` would make the API much clearer.

adr1anh:
[...]

PR #101, opened 2023-11-07
Simplify api

### LINKED ISSUE #63: `supernova::RecursiveSNARK::iter_base_step()` does not iterate `i`
The current setup for `RecursiveSNARK::iter_base_step()` does not iterate `RecursiveSNARK`'s internal step counter `i`. This leads to a [some strange logic](https://github.com/lurk-lab/arecibo/blob/c657b04a005182cb7bc1937f22b70dde747c1bdc/src/supernova/mod.rs#L554) for the first invocation of `prove_step()` that could potentially lead to some unexpected behavior/off-by-one errors.

I believe it should be possible to get rid of the above code snippet and retain logical consistency if `i` is iterated in the first step.

It may also be worth taking the time to rename `i` to something more clear like `num_steps`.

porcuquine:
Closes #63 & #64.

This PR simplifes the SuperNova API by:
- renaming `iter_base_step` to `new`
- minimize args to `supernova::RecursiveSnark::new` — almost matching those of `nova::RecursiveSnark::new`.
  - (The remaining extra arg is new — NonUniformCircuit, which helps derive removed args but cannot itself be eliminated.)
- minimize args to `supernova::RecursiveArgs::prove_step` — matching those of `nova::RecursiveSnark::prove_step`.

### DIFF
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -925,11 +925,6 @@ where
 
     Ok((self.zi_primary.clone(), self.zi_secondary.clone()))
   }
-
-  /// get program counter
-  pub fn get_program_counter(&self) -> G1::Scalar {
-    self.program_counter
-  }
 }
 
 /// SuperNova helper trait, for implementors that provide sets of sub-circuits to be proved via NIVC. `C1` must be a
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -457,7 +457,7 @@ where
     let circuit_primary: SuperNovaAugmentedCircuit<'_, G2, C1> = SuperNovaAugmentedCircuit::new(
       &pp.augmented_circuit_params_primary,
       Some(inputs_primary),
-      &c_primary,
+      c_primary,
       pp.ro_consts_circuit_primary.clone(),
       num_augmented_circuits,
     );
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -939,11 +939,6 @@ where
   C1: StepCircuit<G1::Scalar>,
   C2: StepCircuit<G2::Scalar>,
 {
-  /// Initial program counter is the initial circuit index as a `Scalar`.
-  fn initial_program_counter(&self) -> G1::Scalar {
-    G1::Scalar::from(self.initial_circuit_index() as u64)
-  }
-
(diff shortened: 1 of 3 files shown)


src/supernova/mod.rs
```diff
@@ -577,10 +581,10 @@ where
   pub fn prove_step<C1: StepCircuit<G1::Scalar>, C2: StepCircuit<G2::Scalar>>(
     &mut self,
     pp: &PublicParams<G1, G2, C1, C2>,
-    circuit_index: usize,
     c_primary: &C1,
     c_secondary: &C2,
   ) -> Result<(), SuperNovaError> {
+    let circuit_index = c_primary.circuit_index();
```
adr1anh, line 587: Consider adding a check to ensure this matches `self.program_counter`.

adr1anh (commented):
Does it still make sense for `RecursiveSNARK` to expose `[get_program_counter](https://github.com/lurk-lab/arecibo/blob/8e7509459f2dd51c83f9199186c8db30d5bf2a6d/src/supernova/mod.rs#L926)`

src/supernova/mod.rs
```diff
@@ -938,9 +942,14 @@ where
...
-  /// Initial program counter, defaults to zero.
+  /// Initial program counter is the initial circuit index as a `Scalar`.
   fn initial_program_counter(&self) -> G1::Scalar {
-    G1::Scalar::ZERO
+    G1::Scalar::from(self.initial_circuit_index() as u64)
+  }
+
+  /// Initial circuit index, defaults to zero.
+  fn initial_circuit_index(&self) -> usize {
+    0
```
huitseeker, line 952: This API allows de-synchronizing the `usize` and `Scalar` representations of the initial program counter. It would be better to add `initial_program_counter` as an extension trait (a trait with a blanket implementation for `T: NonUniformCircuit`), then Rust will guarantee *that* implementation is the only one that can apply.

src/supernova/mod.rs
porcuquine: Done.

src/supernova/mod.rs
porcuquine: Done.

huitseeker (approved):
Makes sense to me!

mpenciak (approved):

PR #107, opened 2023-11-09
chore: switch PR bench runners to Buildjet

samuelburnham:
Now that Buildjet has been battle tested for a few weeks, the GitHub-hosted larger runners are removed to simplify the org-wide runner config.


huitseeker (approved):
:vulcan_salute:

PR #108, opened 2023-11-09
Supernova `CompressedSNARK`

mpenciak:


porcuquine:
We should add a benchmark for this.

adr1anh:
Given we have the chance to define our own traits for `RelaxedR1CSSNARK`, I think we should take the opportunity to make the API adhere to [`C-CALLER-CONTROL`](https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control). 
In particular, the `prove` function has to make copies of its inputs, which should instead be cloned by the caller:

```rust
RelaxedR1CSSNARKTrait<G>::prove(
        ck: &CommitmentKey<G>,
        pk: &Self::ProverKey,
        S: Vec<R1CSShape<G>>,
        U: Vec<RelaxedR1CSInstance<G>>,
        W: Vec<RelaxedR1CSWitness<G>>,
    )
```

adr1anh:
Moreover, we should think about whether we want to implement `RelaxedR1CSSNARK` in terms of `BatchedRelaxedR1CSSNARK`.

adr1anh (commented):
Some things left to figure out:

**Embedding number of instances**: 

Variable-length-message protocols can be tricky with Fiat-Shamir. We may want to embed the number of "batches" at the transcript initialization to prevent a transcript from being replayed.

**Sending batches of values**:
In rounds where we send a slice of values for each instance, we are currently calling `absorb` on each slice, and each call writes the domain separator. This is *slightly* inefficient, but still correct. Does it make more sense to just absorb the concatenation of all values?

src/spartan/polys/multilinear.rs
```diff
@@ -64,10 +64,14 @@ impl<Scalar: PrimeField> MultilinearPolynomial<Scalar> {
...
 
-  /// Bounds the polynomial's top variable using the given scalar.
+  /// Binds the polynomial's top variable using the given scalar.
   ///
   /// This operation modifies the polynomial in-place.
-  pub fn bound_poly_var_top(&mut self, r: &Scalar) {
+  pub fn bind_poly_var_top(&mut self, r: &Scalar) {
+    if self.num_vars == 0 {
+      return;
+    }
```
porcuquine, line 73: This seems like it accommodates a mistaken usage. Do we actually need this to be a silent no-op, or can we make this an assertion?

src/spartan/sumcheck.rs
```diff
@@ -360,4 +360,94 @@ impl<G: Group> SumcheckProof<G> {
...
+    comb_func: F,
+    transcript: &mut G::TE,
+  ) -> Result<(Self, Vec<G::Scalar>, G::Scalar, Vec<Vec<G::Scalar>>), NovaError>
+  where
+    F: Fn(&G::Scalar, &G::Scalar, &G::Scalar, &G::Scalar) -> G::Scalar + Sync,
+  {
+    let mut r: Vec<G::Scalar> = Vec::new();
+    let mut polys: Vec<CompressedUniPoly<G::Scalar>> = Vec::new();
+    let mut claim_per_round = *claim;
+
```
porcuquine, line 381: I think there should be a check that the vectors and coefficients are all of the same length. If that's not the case, this is going to have unexpected results — which I don't think we want to take responsibility for producing.

huitseeker:
Closing in favor of #131

PR #110, opened 2023-11-09
Avoid recreating default R1CS instance

### LINKED ISSUE #109: SuperNova: Duplicated variables
In [`SuperNovaAugmentedCircuit::synthesize_base_case`](https://github.com/lurk-lab/arecibo/blob/4d2820ab4d7326ac655d9a0573c4584393dea29c/src/supernova/circuit.rs#L290-L294), the `default` R1CS is allocated at every iteration. Placing it outside the loop would avoid duplicated variables and constraints.

adr1anh:
Fixes #109


huitseeker (approved):
🙏

PR #112, opened 2023-11-09
SuperNova: clean-up `RecursiveSNARK`

adr1anh:
This PR makes some small changes to the `RecursiveSNARK` struct.

- Ensures that synthesizing the circuit during public parameter generation does not fail.
- Document fields of `RelaxedSNARK`
- Convert `r_W_secondary` and `r_U_secondary` from `Vec<Option<T>>` to `T` since we are always only folding a single circuit on the secondary curve.

### DIFF
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -214,7 +214,8 @@ where
           num_circuits,
         );
         let mut cs: ShapeCS<G1> = ShapeCS::new();
-        let _ = circuit_primary.synthesize(&mut cs);
+        let res = circuit_primary.synthesize(&mut cs);
+        assert!(res.is_ok(), "circuit synthesis failed");
 
         // We use the largest commitment_key for all instances
         let r1cs_shape_primary = cs.r1cs_shape();
@@ -239,7 +240,8 @@ where
       num_circuits,
     );
     let mut cs: ShapeCS<G2> = ShapeCS::new();
-    let _ = circuit_secondary.synthesize(&mut cs);
+    let res = circuit_secondary.synthesize(&mut cs);
+    assert!(res.is_ok(), "circuit synthesis failed");
     let (r1cs_shape_secondary, ck_secondary) = cs.r1cs_shape_and_key(ck_hint2);
     let circuit_shape_secondary = CircuitShape::new(r1cs_shape_secondary, F_arity_secondary);
 
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -214,8 +214,9 @@ where
           num_circuits,
         );
         let mut cs: ShapeCS<G1> = ShapeCS::new();
-        let res = circuit_primary.synthesize(&mut cs);
-        assert!(res.is_ok(), "circuit synthesis failed");
+        let _ = circuit_primary
+          .synthesize(&mut cs)
+          .expect("circuit synthesis failed");
 
         // We use the largest commitment_key for all instances
         let r1cs_shape_primary = cs.r1cs_shape();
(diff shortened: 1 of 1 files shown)


src/supernova/mod.rs
```diff
@@ -214,7 +214,8 @@ where
           num_circuits,
         );
         let mut cs: ShapeCS<G1> = ShapeCS::new();
-        let _ = circuit_primary.synthesize(&mut cs);
+        let res = circuit_primary.synthesize(&mut cs);
+        assert!(res.is_ok(), "circuit synthesis failed");
```
huitseeker, line 218: aka `circuit_primary.synthesize(&mut cs).expect("circuit synthesis failed");`

src/supernova/mod.rs
```diff
@@ -239,7 +240,8 @@ where
       num_circuits,
     );
     let mut cs: ShapeCS<G2> = ShapeCS::new();
-    let _ = circuit_secondary.synthesize(&mut cs);
+    let res = circuit_secondary.synthesize(&mut cs);
+    assert!(res.is_ok(), "circuit synthesis failed");
```
huitseeker, line 244: aka aka `circuit_secondary.synthesize(&mut cs).expect("circuit synthesis failed");`

src/supernova/mod.rs
```diff
@@ -446,12 +459,12 @@ where
         scalar_as_base::<G1>(pp.digest()),
         G1::Scalar::ZERO,
         z0_primary,
-        None,
-        None,
-        None,
-        None,
-        Some(G1::Scalar::from(circuit_index as u64)),
-        G1::Scalar::ZERO, // set augmented circuit index selector to 0 in base case
+        None,                                         // zi = None for basecase
```
huitseeker, line 462: Weird spacing.

src/supernova/mod.rs
```diff
@@ -486,12 +499,12 @@ where
         pp.digest(),
         G2::Scalar::ZERO,
         z0_secondary,
-        None,
-        None,
-        Some(&u_primary),
-        None,
-        None,
-        G2::Scalar::from(circuit_index as u64),
+        None,                                   // zi = None
```
huitseeker, line 502: Nit: weird spacing

src/supernova/mod.rs
```diff
@@ -794,18 +798,14 @@ where
...
-        Ok(())
-      }
-    })?;
+    if self.r_U_secondary.X.len() != 2 {
+      debug!(
+        "r_U_secondary got instance length {:?} != {:?}",
+        self.r_U_secondary.X.len(),
+        2
+      );
+      Err(SuperNovaError::NovaError(NovaError::ProofVerifyError))?
```
huitseeker, line 807: I think this is missing an important `return` statement.

huitseeker (commented):
I have my issues on the over-use of option for values to be eventually assigned once, but I'll keep them to myself until another time. Nice improvement, left comments inline.

src/supernova/mod.rs
adr1anh: I thought the `?` at the end would work too, but the return is more clear.

src/supernova/mod.rs
adr1anh: Added an auxiliary variable to make `rustfmt` play nice.

src/supernova/mod.rs
adr1anh: Added an auxiliary variable to make `rustfmt` play nice.

src/supernova/mod.rs
adr1anh: Thanks! I thought I was missing something. Looks nicer like this.

src/supernova/mod.rs
adr1anh: Thanks!

src/supernova/mod.rs
```diff
@@ -214,7 +214,9 @@ where
           num_circuits,
         );
         let mut cs: ShapeCS<G1> = ShapeCS::new();
-        let _ = circuit_primary.synthesize(&mut cs);
+        let _ = circuit_primary
```
huitseeker, line 217: Nit: I don't think you need the `let _ =`.

src/supernova/mod.rs
```diff
@@ -239,7 +241,9 @@ where
       num_circuits,
     );
     let mut cs: ShapeCS<G2> = ShapeCS::new();
-    let _ = circuit_secondary.synthesize(&mut cs);
+    let _ = circuit_secondary
```
huitseeker, line 244: Nit: I don't think you need the `let _ =`.

huitseeker (approved):
Thanks!

huitseeker (approved):

PR #114, opened 2023-11-10
SuperNova: Improved `RecursiveSNARK::verify`

adr1anh:
This PR fixes some issues with `supernova::RecursiveSNARK::verify`:

- Previously, only the last folded primary circuit's instance was being verified. We now verify all all instances
- For circuits which have not been executed, the wrong public parameters were being used to generate a default `RelaxedR1CSInstance` to be hashed into the RO.
- Added extra sanity checks

### DIFF
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -44,12 +44,6 @@ use circuit::{
 
 use self::error::SuperNovaError;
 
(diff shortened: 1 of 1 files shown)


mpenciak (approved):
This fixes an important soundness bug, looks good!

PR #117, opened 2023-11-14
Curve checks on committed point (Nova Forward port)

huitseeker:
ports https://github.com/microsoft/Nova/pull/258

### DIFF
--- a/src/gadgets/ecc.rs
+++ b/src/gadgets/ecc.rs
@@ -63,7 +63,53 @@ where
     Ok(AllocatedPoint { x, y, is_infinity })
   }
 
-  /// Allocates a default point on the curve.
+  /// checks if `self` is on the curve or if it is infinity
+  pub fn check_on_curve<CS>(&self, mut cs: CS) -> Result<(), SynthesisError>
+  where
+    CS: ConstraintSystem<G::Base>,
+  {
+    // check that (x,y) is on the curve if it is not infinity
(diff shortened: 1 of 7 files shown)


src/gadgets/ecc.rs
```diff
@@ -63,7 +63,53 @@ where
...
+      } else {
+        Ok(
+          *x_cube.get_value().get()?
+            + *self.x.get_value().get()? * G::get_curve_params().0
+            + G::get_curve_params().1,
+        )
+      }
+    })?;
+
+    cs.enforce(
```
porcuquine, line 90: Should have a comment summarizing the constraint rather than (only) using the namespace for this.

src/gadgets/ecc.rs
```diff
@@ -63,7 +63,53 @@ where
...
+      |lc| {
+        lc + x_cube.get_variable()
+          + (G::get_curve_params().0, self.x.get_variable())
+          + (G::get_curve_params().1, CS::one())
+      },
+      |lc| lc + CS::one() - self.is_infinity.get_variable(),
+      |lc| lc + rhs.get_variable(),
+    );
+
+    // check that (1-infinity) * y_square = rhs
```
porcuquine, line 101: Prefer to order this the terms the same as the linear combination itself does.

porcuquine (approved):
Approved with minor notes about best practice for constraint comments.

PR #119, opened 2023-11-14
SuperNova: check T is on curve

adr1anh:
Companion to #117

Add check that `T` is on the curve in the SuperNova circuit.


huitseeker (approved):
Excellent!

mpenciak (approved):

PR #120, opened 2023-11-14
SuperNova: Improve circuit

adr1anh:
This PR fixes some minor annoyances in the SuperNova circuit and adds documentation.

- Compute number of RO inputs for public output via a function
- Compute instance selector vector from `last_augmented_circuit_index` in `alloc_witness`
  - Reuse the selector across the base and non-base branches.
- Add comments

### DIFF
--- a/src/gadgets/utils.rs
+++ b/src/gadgets/utils.rs
@@ -70,6 +70,7 @@ pub fn alloc_one<F: PrimeField, CS: ConstraintSystem<F>>(mut cs: CS) -> Allocate
 
 /// alloc a field as a constant
 /// implemented refer from <https://github.com/lurk-lab/lurk-rs/blob/4335fbb3290ed1a1176e29428f7daacb47f8033d/src/circuit/gadgets/data.rs#L387-L402>
+#[allow(unused)]
 pub fn alloc_const<F: PrimeField, CS: ConstraintSystem<F>>(
   mut cs: CS,
   val: F,
--- a/src/gadgets/utils.rs
+++ b/src/gadgets/utils.rs
@@ -198,6 +198,7 @@ pub fn alloc_num_equals<F: PrimeField, CS: ConstraintSystem<F>>(
 }
 
 /// Check that two numbers are equal and return a bit
+#[allow(unused)]
 pub fn alloc_num_equals_const<F: PrimeField, CS: ConstraintSystem<F>>(
   mut cs: CS,
   a: &AllocatedNum<F>,
--- a/src/supernova/circuit.rs
+++ b/src/supernova/circuit.rs
@@ -22,8 +22,7 @@ use crate::{
       AllocatedRelaxedR1CSInstance,
     },
     utils::{
-      alloc_const, alloc_num_equals, alloc_scalar_as_base, alloc_zero, conditionally_select_vec,
-      le_bits_to_num, scalar_as_base,
+      alloc_num_equals, alloc_scalar_as_base, alloc_zero, conditionally_select_vec, le_bits_to_num,
     },
   },
   r1cs::{R1CSInstance, RelaxedR1CSInstance},
@@ -45,7 +44,10 @@ use abomonation_derive::Abomonation;
 use ff::Field;
 use serde::{Deserialize, Serialize};
 
-use super::{num_ro_inputs, utils::get_from_vec_alloc_relaxed_r1cs};
+use super::{
+  num_ro_inputs,
+  utils::{get_from_vec_alloc_relaxed_r1cs, get_selector_vec_from_index},
+};
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Abomonation)]
 pub struct SuperNovaAugmentedCircuitParams {
@@ -72,12 +74,21 @@ impl SuperNovaAugmentedCircuitParams {
 pub struct SuperNovaAugmentedCircuitInputs<'a, G: Group> {
   pp_digest: G::Scalar,
   i: G::Base,
+  // Input to the circuit for the base case
   z0: &'a [G::Base],
+  // Input to the circuit for the non-base case
   zi: Option<&'a [G::Base]>,
+  // List of `RelaxedR1CSInstance`.
+  // `None` if this is the base case.
+  // Elements are `None` if the circuit at that index was not yet executed.
   U: Option<&'a [Option<RelaxedR1CSInstance<G>>]>,
+  // R1CS proof to be folded into U
   u: Option<&'a R1CSInstance<G>>,
+  // Nova folding proof for accumulating u into U[j]
   T: Option<&'a Commitment<G>>,
+  // Index of the current circuit
   program_counter: Option<G::Base>,
+  // Index j of circuit being folded into U[j]
   last_augmented_circuit_index: G::Base,
 }
 
@@ -137,7 +148,8 @@ impl<'a, G: Group, SC: EnforcingStepCircuit<G::Base>> SuperNovaAugmentedCircuit<
     }
   }
 
-  /// Allocate all witnesses and return
+  /// Allocate all witnesses from the augmented function's non-deterministic inputs.
+  /// Optional entries are allocated as their default values.
   fn alloc_witness<CS: ConstraintSystem<<G as Group>::Base>>(
     &self,
     mut cs: CS,
@@ -153,7 +165,7 @@ impl<'a, G: Group, SC: EnforcingStepCircuit<G::Base>> SuperNovaAugmentedCircuit<
       AllocatedR1CSInstance<G>,
       AllocatedPoint<G>,
       Option<AllocatedNum<G::Base>>,
-      AllocatedNum<G::Base>,
+      Vec<Boolean>,
     ),
     SynthesisError,
   > {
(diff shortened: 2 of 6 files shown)


src/supernova/circuit.rs
```diff
@@ -72,12 +74,21 @@ impl SuperNovaAugmentedCircuitParams {
...
+  // `None` if this is the base case.
+  // Elements are `None` if the circuit at that index was not yet executed.
   U: Option<&'a [Option<RelaxedR1CSInstance<G>>]>,
+  // R1CS proof to be folded into U
   u: Option<&'a R1CSInstance<G>>,
+  // Nova folding proof for accumulating u into U[j]
   T: Option<&'a Commitment<G>>,
+  // Index of the current circuit
   program_counter: Option<G::Base>,
+  // Index j of circuit being folded into U[j]
```
arthurpaulino, line 91: These comments can be turned into real attribute docstrings with an extra `/`

adr1anh: That's neat! thanks for the suggestion!

src/supernova/mod.rs
```diff
@@ -1045,6 +1047,20 @@ pub fn circuit_digest<
   circuit_params.digest()
 }
 
+/// Compute the number of absorbs for the random-oracle computing the circuit output
+/// X = H(vk, i, pc, z0, zi, U)
+fn num_ro_inputs(num_circuits: usize, num_limbs: usize, arity: usize, is_primary: bool) -> usize {
```
porcuquine, line 1052: Finally, thank you. This could also reasonably be a method of `SuperNovaAugmentedCircuit` with even fewer arguments.

src/supernova/test.rs
```diff
@@ -651,7 +651,7 @@ fn test_recursive_circuit() {
   let ro_consts1: ROConstantsCircuit<G2> = PoseidonConstantsCircuit::default();
   let ro_consts2: ROConstantsCircuit<G1> = PoseidonConstantsCircuit::default();
 
-  test_recursive_circuit_with::<G1, G2>(&params1, &params2, ro_consts1, ro_consts2, 9845, 12036);
+  test_recursive_circuit_with::<G1, G2>(&params1, &params2, ro_consts1, ro_consts2, 9841, 12028);
```
porcuquine, line 654: Thanks for getting the last-mile optimization one.

src/supernova/utils.rs
```diff
@@ -23,51 +23,69 @@ use crate::{
...
+  mut cs: CS,
+  target_index: &AllocatedNum<F>,
+  num_indices: usize,
+) -> Result<Vec<Boolean>, SynthesisError> {
+  assert_ne!(num_indices, 0);
+
+  // Trivial case, return [1]
+  if num_indices == 1 {
+    return Ok(vec![Boolean::Constant(true)]);
+  }
```
porcuquine, line 70: I'm not sure this should be a special case. We still need to ensure `target_index == 0` here, right?

src/supernova/utils.rs
```diff
@@ -117,4 +137,32 @@ mod test {
...
+  #[test]
+  fn test_get_selector() {
+    for n in 1..4 {
+      for selected in 0..n {
+        let mut cs = TestConstraintSystem::<Base>::new();
+
+        let allocated_target =
+          alloc_const(&mut cs.namespace(|| "target"), Base::from(selected as u64)).unwrap();
+
+        let selector_vec = get_selector_vec_from_index(&mut cs, &allocated_target, n).unwrap();
```
porcuquine, line 150: I don't think this will ever allocate a target that is out of range, so this test is misleading. The condition `selected < n` will always be truce, since selected is always taken from `0..n`.

porcuquine (changes requested):
I left a couple comments about the possibility of an out-of-range index when creating selectors of length one, and about the test that would have caught it.

src/supernova/mod.rs
adr1anh: I wanted to do that originally, but the function is used by `RecursiveSNARK::verify`, where we don't create a `SuperNovaAugmentedCircuit` instance.

src/supernova/utils.rs
adr1anh: I think you're referring to the test below. I'm extending the range of `selected` to `2*n` to catch the unsatisfiable cases. 
I've also changed the `is_valid` condition to only run in the satisfying case, and checking that the correct bit is set.

src/supernova/utils.rs
adr1anh: For the correctness of this function, I agree that we should then enforce `target_index == 0`.

However, as it is used in the circuit on the primary curve where there is only one running instance, we could entirely remove the allocation of `last_augmented_circuit_index` (make it optional) and use `[Boolean::Constant(true)]` as the selector without the additional constraint.

porcuquine: Sure, I think they are orthogonal issues. In general, there is a tradeoff between complexity and constraint minimization. In cases where we're talking about saving one or two constraints at the cost of additional machinery and code paths, I would think hard before making that choice. The point I'm making is that if one were to choose the policy of *always* seeking minimal constraints, without regard for code complexity, then it's a safe bet that the result will be overwhelming. Let's take up the question of whether further optimization is warranted in a separate issue.

src/supernova/utils.rs
```diff
@@ -23,60 +22,99 @@ use crate::{
...
+  assert_ne!(num_indices, 0);
+
+  // Trivial case: return [1], enforce `target_index == 0`
+  if num_indices == 1 {
+    cs.enforce(
+      || "target_index = 0",
+      |lc| lc,
+      |lc| lc,
+      |lc| lc + target_index.get_variable(),
+    );
```
porcuquine, line 72: I think this is redundant. The general case will handle this one correctly, saving a constraint — but more importantly keeping the logic simple.

src/gadgets/utils.rs
```diff
@@ -197,6 +198,7 @@ pub fn alloc_num_equals<F: PrimeField, CS: ConstraintSystem<F>>(
 }
 
 /// Check that two numbers are equal and return a bit
+#[allow(unused)]
```
huitseeker, line 201: Same comment as for alloc_const.

src/gadgets/utils.rs
```diff
@@ -70,6 +70,7 @@ pub fn alloc_one<F: PrimeField, CS: ConstraintSystem<F>>(mut cs: CS) -> Allocate
 
 /// alloc a field as a constant
 /// implemented refer from <https://github.com/lurk-lab/lurk-rs/blob/4335fbb3290ed1a1176e29428f7daacb47f8033d/src/circuit/gadgets/data.rs#L387-L402>
+#[allow(unused)]
```
huitseeker, line 73: We should bin it I think, this is a Lurk addition not in Nova main. And we have copies of it lying around.

src/supernova/utils.rs
adr1anh: That's fair, I removed it in the latest commit

src/gadgets/utils.rs
adr1anh: Removed, as well as the test below

src/gadgets/utils.rs
adr1anh: Removed it, but the function was used in tests. The commit replaces their usages with other existing functions

adr1anh (commented):
Thank you all for the useful comments, I hope this adresses everything. I think it needs a rebase, so I'll wait for confirmation that all points have been adressed.

adr1anh:
On second thought, I think removing the unused functions should be handled by another PR. They are used by the tests, namely those using `TestROM`. For now, I'd suggest reverting the latest commit and filing an issue to address the dead code separately. 

The fact that the latest commit failed worries me a bit. Replacing the bit computation with the new selector gadget should not have changed the behavior. 
The failure happens in the last iteration, since the `program_counter` is the next one computed by the function, and it does not belong in the ROM. Therefore, I'm not entirely convinced the circuit as a whole is correct, at least as it was before the latest change. For instance, I see that 
[constrain_augmented_circuit_index](https://github.com/adr1anh/arecibo/blob/cc2ccf2d88f95dcdf6899fce74d655d1197034ef/src/supernova/test.rs#L239-L246) is called in `SquareCircuit` but not in `CubicCircuit`.
While it is only there for testing purposes, I'm worried someone might copy it for their own circuit. Do you think a more thorough audit of this circuit is warranted ?

huitseeker (approved):
LGTM!

porcuquine:
> On second thought, I think removing the unused functions should be handled by another PR. They are used by the tests, namely those using `TestROM`. For now, I'd suggest reverting the latest commit and filing an issue to address the dead code separately. WDYT @huitseeker ?
> 
> The fact that the latest commit failed worries me a bit. Replacing the bit computation with the new selector gadget should not have changed the behavior. The failure happens in the last iteration, since the `program_counter` is the next one computed by the function, and it does not belong in the ROM. Therefore, I'm not entirely convinced the circuit as a whole is correct, at least as it was before the latest change. For instance, I see that [constrain_augmented_circuit_index](https://github.com/adr1anh/arecibo/blob/cc2ccf2d88f95dcdf6899fce74d655d1197034ef/src/supernova/test.rs#L239-L246) is called in `SquareCircuit` but not in `CubicCircuit`. While it is only there for testing purposes, I'm worried someone might copy it for their own circuit. Do you think a more thorough audit of this circuit is warranted ?

I agree with your analysis.

porcuquine (approved):
Great work.

PR #122, opened 2023-11-15
refactor: Remove nedless pass-by-value instances, turn on clippy

huitseeker:
- Updated function parameters in `ppsnark.rs` and `snark.rs` from `Vec<G::Scalar>` to `&[G::Scalar]` (introduced in #106),
- Modified the `prove` function in `ppsnark.rs` to also convert `T_row`, `W_row`, `T_col`, and `W_col` from `Vec` to slices (`&[G::Scalar]`).
- Enhanced the `xclippy` alias in `.cargo/config` by adding `-Wclippy::needless_pass_by_value`, and `-Wclippy::unnecessary_mut_passed` and reorganizing its elements.


adr1anh (commented):
With regards to `u_vec` and `w_vec`, I think the right approach here is to make `PolyEval{Instance, Witness}::pad` accept a `Vec` as input rather than a slice to avoid creating copies.

adr1anh (approved):

adr1anh (approved):

PR #127, opened 2023-11-16
SuperNova: Fix ROM test and remove dead code

adr1anh:
The ROM test was broken and not actually doing what it was supposed to. 
In particular:
- The input rom_index was initialized with the first opcode, rather than 0
- `SquareCircuit::synthesize` was checking the next opcode twice
- `constrain_augmented_circuit_index` was not checking what it was supposed to.

This issue was found while trying to remove dead code in PR #120.

### DIFF
--- a/src/supernova/test.rs
+++ b/src/supernova/test.rs
@@ -1,8 +1,4 @@
-use crate::bellpepper::test_shape_cs::TestShapeCS;
-use crate::gadgets::utils::alloc_const;
-use crate::gadgets::utils::alloc_num_equals;
-use crate::gadgets::utils::conditionally_select;
-use crate::gadgets::utils::{alloc_one, alloc_zero};
+use crate::gadgets::utils::alloc_zero;
 use crate::provider::bn256_grumpkin::bn256;
 use crate::provider::bn256_grumpkin::grumpkin;
 use crate::provider::poseidon::PoseidonConstantsCircuit;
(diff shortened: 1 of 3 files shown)


porcuquine:
Let's merge #120 then rebase this to simplify review, please.

adr1anh:
> Let's merge #120 then rebase this to simplify review, please.

Done, the changes are much smaller now.

src/supernova/test.rs
```diff
@@ -466,55 +403,37 @@ where
...
+        &z0_secondary,
+      )
+      .unwrap()
+    });
 
-    let mut recursive_snark =
-      recursive_snark_option.unwrap_or_else(|| match augmented_circuit_index {
-        OPCODE_0 | OPCODE_1 => RecursiveSNARK::new(
+    let circuit_primary = test_rom.primary_circuit(op_code);
+    let circuit_secondary = test_rom.secondary_circuit();
```
porcuquine, line 420: Maybe move these bindings to the top of the loop and use them in the binding of `recursive_snark` also.

porcuquine (dismissed):
This seems fine. Independent of bug fixes, I agree that simplifying the logic of this test example is useful. I think it exists in its current form for historical reasons, and there is little value in preserving the specific 'ROM' use case beyond the extent to which it also serves as a clear test and example. These changes seem to bring it closer to serving that purpose.

src/supernova/test.rs
```diff
@@ -88,55 +38,58 @@ where
     }
   }
 }
-/// c = a + b where a, b is `AllocatedNum`
-pub fn add_allocated_num<F: PrimeField, CS: ConstraintSystem<F>>(
```
huitseeker, line 92: In case this comes up one day, https://github.com/lurk-lab/bellpepper/blob/main/crates/bellpepper-core/src/gadgets/num.rs#L279-L309

huitseeker (dismissed):

src/supernova/test.rs
adr1anh: Good catch!

PR #130, opened 2023-11-22
feat: fold the minroot_serde example under a feature

### LINKED ISSUE #62: Deduplicate minroot and minroot-serde examples
We're ferrying a `minroot_serde` example since https://github.com/lurk-lab/arecibo/commit/2f544af9286a906fadc1fc9d9e9c535358c8e725, which duplicates the entire `minroot` example to demonstrate the fast (de/)serialization of public parameters.

That's a lot of real estate for a relatively simple demonstration. Could the minroot & minroot_serde examples perhaps be merged, possibly switching the behavior with either a CLI parameter or a feature?

huitseeker:
- Removed `examples/minroot_serde.rs` file, which previously managed the serialization and deserialization of public params demonstrated on the Nova VDF,
- Created and updated several utility functions and data structures in `examples/minroot.rs` to accommodate the prior `minroot_serde` demonstration.
- Introduced the `abomonate` feature,

Fixes #62.


examples/minroot.rs
```diff
@@ -19,7 +19,58 @@ use std::time::Instant;
...
 use tracing_texray::TeXRayLayer;
 
-#[derive(Clone, Debug)]
+#[cfg(feature = "abomonate")]
+mod utils {
+  use super::*;
+  use std::{io::Write, mem::size_of};
+
+  pub const FILEPATH: &str = "/tmp/data";
+  /// Unspeakable horrors from a type safety PoV
```
adr1anh, line 28: 🥲

examples/minroot.rs
```diff
@@ -19,7 +19,58 @@ use std::time::Instant;
...
+  unsafe fn any_as_u8_slice<T: Sized>(p: &T) -> &[u8] {
+    ::core::slice::from_raw_parts((p as *const T) as *const u8, ::core::mem::size_of::<T>())
+  }
+
+  /// this is dangerous
+  #[allow(non_snake_case)]
+  unsafe fn entomb_F<F: PrimeField, W: Write>(f: &F, bytes: &mut W) -> std::io::Result<()> {
+    println!("entomb: {}", size_of::<F>());
```
adr1anh, line 36: Is this `println` intentional?

examples/minroot.rs
```diff
@@ -19,7 +19,58 @@ use std::time::Instant;
...
+  }
+
+  /// this is dangerous
+  #[allow(non_snake_case)]
+  unsafe fn exhume_F<'a, F: PrimeField>(f: &mut F, bytes: &'a mut [u8]) -> Option<&'a mut [u8]> {
+    let (mine, rest) = bytes.split_at_mut(size_of::<F>());
+    let mine = (mine as *const [u8]) as *const F;
+    std::ptr::write(f, std::ptr::read(mine));
+    Some(rest)
+  }
```
adr1anh, line 49: Can this function return `None`?

adr1anh (dismissed):
Scary stuff, but seems good to me.

examples/minroot.rs
huitseeker: Yes: most of the example functions through stdout printing, and in so far as we incorporate the demonstration of the abomonate capability, that's how we do it as well.

examples/minroot.rs
huitseeker: No. The reason it is wrapped in option in this way is that it is meant to emulate the implementation of a trait (Abomonate) that requires this API. Nonetheless, I've added a commit that removes the wrapping.

adr1anh (approved):

PR #131, opened 2023-11-22
Supernova `CompressedSNARK`

huitseeker:
This is #108 re-opened from a feature branch.

### DIFF
--- a/src/spartan/batched.rs
+++ b/src/spartan/batched.rs
@@ -1,0 +1,66 @@
+//! TODO: Doc
+
+use std::marker::PhantomData;
+
+use serde::{Deserialize, Serialize};
+
+use crate::traits::{
+  evaluation::EvaluationEngineTrait,
+  snark::{BatchedRelaxedR1CSSNARKTrait, DigestHelperTrait},
+  Group,
+};
+
+use super::snark::RelaxedR1CSSNARK;
+
+/// TODO: Doc
+#[derive(Clone, Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct ProverKey<G: Group, EE: EvaluationEngineTrait<G>> {
+  _p: PhantomData<(G, EE)>,
+}
+
+/// TODO: Doc
+#[derive(Clone, Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct VerifierKey<G: Group, EE: EvaluationEngineTrait<G>> {
+  _p: PhantomData<(G, EE)>,
+}
+
+impl<G: Group, EE: EvaluationEngineTrait<G>> DigestHelperTrait<G> for VerifierKey<G, EE> {
+  fn digest(&self) -> <G as Group>::Scalar {
+    todo!()
+  }
+}
+
+impl<G: Group, EE: EvaluationEngineTrait<G>> BatchedRelaxedR1CSSNARKTrait<G>
+  for RelaxedR1CSSNARK<G, EE>
+{
+  type ProverKey = ProverKey<G, EE>;
+
+  type VerifierKey = VerifierKey<G, EE>;
+
+  fn setup(
+    ck: &crate::CommitmentKey<G>,
+    S: &[crate::r1cs::R1CSShape<G>],
+  ) -> Result<(Self::ProverKey, Self::VerifierKey), crate::errors::NovaError> {
+    todo!()
+  }
+
+  fn prove(
+    ck: &crate::CommitmentKey<G>,
+    pk: &Self::ProverKey,
+    S: &[crate::r1cs::R1CSShape<G>],
+    U: &[crate::r1cs::RelaxedR1CSInstance<G>],
+    W: &[crate::r1cs::RelaxedR1CSWitness<G>],
+  ) -> Result<Self, crate::errors::NovaError> {
+    todo!()
+  }
+
+  fn verify(
+    &self,
+    vk: &Self::VerifierKey,
+    U: &[crate::r1cs::RelaxedR1CSInstance<G>],
+  ) -> Result<(), crate::errors::NovaError> {
+    todo!()
+  }
+}
--- a/src/spartan/batched.rs
+++ b/src/spartan/batched.rs
@@ -2,13 +2,23 @@
 
 use std::marker::PhantomData;
 
+use ff::Field;
 use serde::{Deserialize, Serialize};
 
+use super::{
+  compute_eval_table_sparse,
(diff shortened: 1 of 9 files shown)


src/traits/snark.rs
```diff
@@ -61,6 +62,44 @@ pub trait RelaxedR1CSSNARKTrait<G: Group>:
...
+  type VerifierKey: Send + Sync + Serialize + for<'de> Deserialize<'de> + DigestHelperTrait<G>;
+  //+ Abomonation; // TODO: Uncomment this
+
+  /// This associated function (not a method) provides a hint that offers
+  /// a minimum sizing cue for the commitment key used by this SNARK
+  /// implementation. The commitment key passed in setup should then
+  /// be at least as large as this hint.
+  fn commitment_key_floor() -> Box<dyn for<'a> Fn(&'a R1CSShape<G>) -> usize> {
+    default_commitment_key_hint()
+  }
```
adr1anh, line 82: This will change to `ck_floor` to match upstream

src/traits/snark.rs
```diff
@@ -61,6 +62,44 @@ pub trait RelaxedR1CSSNARKTrait<G: Group>:
...
+    S: &[R1CSShape<G>],
+  ) -> Result<(Self::ProverKey, Self::VerifierKey), NovaError>;
+
+  /// Produces a new SNARK for a batch of relaxed R1CS
+  fn prove(
+    ck: &CommitmentKey<G>,
+    pk: &Self::ProverKey,
+    S: &[R1CSShape<G>],
+    U: &[RelaxedR1CSInstance<G>],
+    W: &[RelaxedR1CSWitness<G>],
```
adr1anh, line 96: Can we make this take the inputs by value? Maybe even `impl IntoIterator<Item = ...>`?

adr1anh:
I think the order of merge should be #140 and then #136. The former has some factored out code snippets which are copied from `batched_ppsnark`, so I can update #136 to use these.

src/spartan/batched.rs
```diff
@@ -0,0 +1,645 @@
...
+};
+
+/// A succinct proof of knowledge of a witness to a batch of relaxed R1CS instances
+/// The proof is produced using Spartan's combination of the sum-check and
+/// the commitment to a vector viewed as a polynomial commitment
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct BatchedRelaxedR1CSSNARK<G: Group, EE: EvaluationEngineTrait<G>> {
+  sc_proof_outer: SumcheckProof<G>,
+  // Claims ([Azᵢ(τᵢ), ])
```
porcuquine, line 48: Is this complete?

src/spartan/batched.rs
```diff
@@ -0,0 +1,645 @@
...
+      .zip(U.iter())
+      .map(|(w, u)| [w.clone(), vec![u.u], u.X.clone()].concat())
+      .collect::<Vec<Vec<_>>>();
+
+    let (num_rounds_x, num_rounds_y): (Vec<_>, Vec<_>) = S
+      .iter()
+      .map(|s| (s.num_cons.log_2(), s.num_vars.log_2() + 1))
+      .unzip();
+    let num_rounds_x_max = *num_rounds_x.iter().max().unwrap();
+    let num_rounds_y_max = *num_rounds_y.iter().max().unwrap();
```
porcuquine, line 178: You could do this all in one pass with `fold`.

src/spartan/batched.rs
```diff
@@ -0,0 +1,645 @@
...
+
+    // Generate tau polynomial corresponding to eq(τ, τ², τ⁴ , …)
+    // for a random challenge τ
+    let tau = transcript.squeeze(b"t")?;
+    let polys_tau = num_rounds_x
+      .iter()
+      .map(|&num_rounds_x| PowPolynomial::new(&tau, num_rounds_x).evals())
+      .map(MultilinearPolynomial::new)
+      .collect::<Vec<_>>();
```
porcuquine, line 187: I follow this, it is extremely wasteful because you're computing all the prefix powers over in every iterations. It would be better to compute all the powers once, then create a `PowPolynomial` per power — or even just add a `PowPolynomial::new_up_to()` (or whatever).

porcuquine: I'll make a PR for this.

porcuquine: #148

src/spartan/batched.rs
```diff
@@ -0,0 +1,645 @@
...
+    let (claim_outer_final, r_x) = self.sc_proof_outer.verify_batch(
+      &vec![G::Scalar::ZERO; num_instances],
+      &num_rounds_x,
+      &outer_r_powers,
+      3,
+      &mut transcript,
+    )?;
+
+    // Collect evaluation point for each instance
+    let r_x = num_rounds_x
```
porcuquine, line 449: A little more documentation of exactly what's going on here would not hurt.

src/spartan/polys/power.rs
```diff
@@ -10,26 +11,35 @@ use ff::PrimeField;
...
     }
   }
 
+  /// Create powers the following powers of `t`:
+  /// [t^{2^0}, t^{2^1}, ..., t^{2^{ell-1}}]
+  pub(crate) fn powers(t: &Scalar, ell: usize) -> Vec<Scalar> {
+    successors(Some(*t), |p: &Scalar| Some(p.square()))
+      .take(ell)
+      .collect::<Vec<_>>()
+  }
```
adr1anh, line 34: Small nit, but there is already a `powers` function in `spartan/mod.rs`. Maybe this one could be named `squared_powers?`

src/spartan/batched.rs
porcuquine: Sorry for late response, and I'm not particularly saying this is necessary. However, I just meant that you could still do that in one pass with fold. The accumulated value could be `(Vec<usize>, Vec<usize>, usize, usize)` and the function to fold would be something like (untested):
```
|acc, s| {
    let num_rounds_x = s.num_cons.log_2();
    let num_rounds_y = s.num_vars.log_2();
    (num_rounds_x, num_rounds_y, num_rounds_x.max(acc.2), num_rounds_y.max(acc.3))
}
```

src/spartan/batched.rs
huitseeker: https://hackmd.io/@adr1anh/BJw1g0aBT#Non-uniform-instances

adr1anh (approved):

mpenciak (approved):

Issue #75, opened 2023-10-18
Acceleration of SpMVM in folding


huitseeker:
### Background:
Both the Nova project and our for Arecibo consider performance of folding to be critical. During our recent analysis, we noticed that the commitment to cross-terms is a significant component of the folding performance. This observation holds true even when the costs are, for large enough ops, primarily dominated by an MSM operation, as predicted by theoretical analysis.

### Findings:

The texray graph showcasing our investigation's results is as follows:
```
RecursiveSNARK::prove_step                       5s     ├───────────────────────────────────────────────────────────────┤
    <MultiFrame as StepCircuit>::synthesize      1s     ├────────────┤
    <_ as Group>::vartime_multiscalar_mul      947ms                  ├─────────┤
    NIFS::prove                                  3s                              ├─────────────────────────────────────┤
        AZ_1, BZ_1, CZ_1                         1s                              ├─────────────┤
        AZ_2, BZ_2, CZ_2                       767ms                                             ├───────┤
        cross terms                            263ms                                                      ├─┤
        T                                      202ms                                                         ├┤
        <_ as Group>::vartime_multiscalar_mul  674ms                                                           ├──────┤
    <_ as Group>::vartime_multiscalar_mul        5ms                                                                    ┆
```
@winston-h-zhang to provide more information on reproduction here

- The commitment to cross-terms is evident at the following location in the codebase:
[arecibo/src/nifs.rs#L52C1-L53](https://github.com/lurk-lab/arecibo/blob/83982b3babe714d9d1370cbdea1c91945062c31c/src/nifs.rs#L52C1-L53)

- The primary computational complexity in this function arises when computing a matrix-vector product between the R1CS matrices and an incoming instance-witness pair. This is visible at:
[arecibo/src/r1cs/mod.rs#L285-L301](https://github.com/lurk-lab/arecibo/blob/81320d71d050e8d1f4e4580cef269596b8623376/src/r1cs/mod.rs#L285-L301)

- The matrix-vector product terms are represented as AZ_1, BZ_1, CZ_1 and AZ_2, BZ_2, CZ_2 in the texray graph.

- To optimize the operation's speed, [we transitioned the matrix representation from COO to CSR](https://github.com/lurk-lab/arecibo/pull/38) given that they are represented in sparse form.

### Challenges:

[...]

PR #140, opened 2023-11-27
Fix Batched SNARK

adr1anh:
- Rework all batched Sumcheck provers to handle instances of different sizes
- Add `SumcheckProof::verify_batch` to handle scaling. 
- Improve `batch_verify` of MLE evaluations to also handle different sized claims -> should help existing SNARK as well.

### DIFF
--- a/src/spartan/batched.rs
+++ b/src/spartan/batched.rs
@@ -44,6 +44,7 @@ use crate::{
 #[serde(bound = "")]
 pub struct BatchedRelaxedR1CSSNARK<G: Group, EE: EvaluationEngineTrait<G>> {
   sc_proof_outer: SumcheckProof<G>,
+  // Claims ([Azᵢ(τᵢ), ])
   claims_outer: (Vec<G::Scalar>, Vec<G::Scalar>, Vec<G::Scalar>),
   evals_E: Vec<G::Scalar>,
   sc_proof_inner: SumcheckProof<G>,
@@ -249,20 +250,11 @@ impl<G: Group, EE: EvaluationEngineTrait<G>> BatchedRelaxedR1CSSNARKTrait<G>
       comb_func_outer,
       &mut transcript,
     )?;
-    assert_eq!(num_rounds_x_max, r_x.len());
 
     let r_x = num_rounds_x
       .iter()
       .map(|&num_rounds| r_x[(num_rounds_x_max - num_rounds)..].to_vec())
       .collect::<Vec<_>>();
-
-    // DEBUG
-    for ((eval, poly), r_x) in claims_outer[1].iter().zip(polys_Az.iter()).zip(r_x.iter()) {
-      assert_eq!(*eval, MultilinearPolynomial::evaluate_with(poly, r_x));
-    }
-    for ((eval, poly), r_x) in claims_outer[2].iter().zip(polys_Bz.iter()).zip(r_x.iter()) {
-      assert_eq!(*eval, MultilinearPolynomial::evaluate_with(poly, r_x));
-    }
 
     // Extract evaluations of Az, Bz from Sumcheck and Cz, E at r_x
     let (evals_Az_Bz_Cz, evals_E): (Vec<_>, Vec<_>) = claims_outer[1]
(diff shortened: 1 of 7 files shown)


src/spartan/batched.rs
```diff
@@ -254,106 +222,103 @@ impl<G: Group, EE: EvaluationEngineTrait<G>> BatchedRelaxedR1CSSNARKTrait<G>
...
-      &mut vec_poly_Bz,
-      &mut vec_poly_uCz_E,
+    // Verify outer sumcheck: Az * Bz - uCz_E for each instance
+    let (sc_proof_outer, r_x, claims_outer) = SumcheckProof::prove_cubic_with_additive_term_batch(
+      &vec![G::Scalar::ZERO; num_instances],
+      &num_rounds_x,
+      polys_tau
+        .iter()
+        .map(PowPolynomial::evals)
+        .map(MultilinearPolynomial::new)
```
huitseeker, line 232: We could merge this in a single map.

src/spartan/batched.rs
```diff
@@ -254,106 +222,103 @@ impl<G: Group, EE: EvaluationEngineTrait<G>> BatchedRelaxedR1CSSNARKTrait<G>
...
-        .zip(evals_Bs)
-        .zip(evals_Cs)
-        .map(|((eval_A, eval_B), eval_C)| MultilinearPolynomial::new(inner(eval_A, eval_B, eval_C)))
+      S.par_iter()
+        .zip(r_x.par_iter())
+        .map(|(s, r_x)| {
+          let evals_rx = EqPolynomial::new(r_x.clone()).evals();
+          let (eval_A, eval_B, eval_C) = compute_eval_table_sparse(s, &evals_rx);
+          MultilinearPolynomial::new(inner(eval_A, eval_B, eval_C))
+        })
```
huitseeker, line 313: Consider instead 
```rust
    S.par_iter()
        .zip(r_x.into_par_iter())
        .map(|(s, r_x)| {
          let evals_rx = EqPolynomial::new(r_x).evals();
          let (eval_A, eval_B, eval_C) = compute_eval_table_sparse(s, &evals_rx);
          MultilinearPolynomial::new(inner(eval_A, eval_B, eval_C))
        })
        .collect::<Vec<_>>()
```

src/spartan/mod.rs
```diff
@@ -38,29 +38,43 @@ pub struct PolyEvalWitness<G: Group> {
...
-      Vec::new()
-    }
-  }
+  /// Given [Pᵢ] and [sᵢ], compute P = ∑ᵢ sᵢ⋅Pᵢ
+  ///
+  /// # Details
+  ///
+  /// We allow the input polynomials to have different sizes, and interpret smaller ones as
+  /// being padded with 0 to the maximum size of all polynomials.
+  fn batch_diff_size(W: Vec<PolyEvalWitness<G>>, s: G::Scalar) -> PolyEvalWitness<G> {
```
huitseeker, line 47: You can probably get away with the following signature:
```rust
  fn batch_diff_size(W: &mut [PolyEvalWitness<G>], s: G::Scalar) -> PolyEvalWitness<G> 
```

huitseeker (commented):
This is directionally great, I've left a bunch of comments inline.

I'm afraid CI isn't running for me on this, I get a compilation error:
```
error[E0308]: mismatched types
   --> src/spartan/snark.rs:514:57
    |
514 |   let w_joint = PolyEvalWitness::batch_diff_size(w_vec, powers_of_gamma);
    |                 --------------------------------        ^^^^^^^^^^^^^^^ expected associated type, found `Vec<<G as Group>::Scalar>`
    |                 |
    |                 arguments to this function are incorrect
    |
    = note: expected associated type `<G as traits::Group>::Scalar`
                        found struct `Vec<<G as traits::Group>::Scalar>`
note: associated function defined here
   --> src/spartan/mod.rs:47:6
    |
47  |   fn batch_diff_size(W: Vec<PolyEvalWitness<G>>, s: G::Scalar) -> PolyEvalWitness<G> {
    |      ^^^^^^^^^^^^^^^                             ------------

For more information about this error, try `rustc --explain E0308`.
error: could not compile `nova-snark` (lib) due to previous error
```

For the functions `prove_cubic_with_additive_term_batch` and `prove_quad_batch`, it would be great to have micro-benchmarks.
Fast ways to get there:
- using [tracing-texray](https://docs.rs/tracing-texray/latest/tracing_texray/) on comparable examples, as demonstrated in e.g. [this comment](https://github.com/lurk-lab/arecibo/issues/75#issuecomment-1769091911)
- using a quick inline `#{bench]`
- using a fancier bench using [iai](https://crates.io/crates/iai)

src/spartan/snark.rs
```diff
@@ -443,38 +455,43 @@ pub(super) fn batch_eval_prove<G: Group>(
...
-  let num_claims = w_vec_padded.len();
   let powers_of_rho = powers::<G>(&rho, num_claims);
-  let claim_batch_joint = u_vec_padded
-    .iter()
-    .zip(powers_of_rho.iter())
-    .map(|(u, p)| u.e * p)
-    .sum();
 
-  let mut polys_left: Vec<MultilinearPolynomial<G::Scalar>> = w_vec_padded
+  // Create clones of polynomials to be given to Sumcheck
```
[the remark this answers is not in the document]
adr1anh: Thanks for suggesting `multiunzip`, it's really useful in many places in the batched snarks

src/spartan/mod.rs
adr1anh: To be honest, I think the signature should depend on whether we need to use the input polynomials as "scratch" buffer. My intuition is yes, since multiplying the `p_i`s independently by `s_i` first and then summing them has nicer memory access patterns over multiple threads. With the current implementation which overwrites the input polynomials with temporary computation, they are no longer be usable after the function is done. Therefore, I feel like it makes sense to consume them. What are the advantages of taking a `&mut` input?

adr1anh (commented):
Thanks for the great comments, I've adressed everything I could. Let me know what you think

src/spartan/mod.rs
huitseeker: > What are the advantages of taking a &mut input?

Passing the inputs to the computation when someone else owns the allocation you're about to make, admittedly, unusable. It's just a composability / capability thing.

huitseeker (approved):

Issue #137, opened 2023-11-27
Optimizing the memory pipeline


winston-h-zhang:
This is an umbrella issue for optimizing the memory allocations in `arecibo`. The original whole PR is contained in #118 -- however, to have a higher degree of confidence in the breakdown of optimizations, we split the changes into several self-contained PRs. For each of these separate PRs, we provide benchmarks to show that they individually contribute to better performance.

huitseeker:
Note : those changes should be ported to Supernova relatively quickly, as #197 threatens to replace the old (memory-optimized) code paths with the batched (non-memory-optimized) ones, losing the progress we've made here.

huitseeker:
@winston-h-zhang Can we close this after #227 ?

PR #144, opened 2023-11-28
Remove large vector allocations

winston-h-zhang:
Addresses #137.

This PR addresses the cloning/allocation of large vectors used in the intermediary steps of each Nova `prove_step` computation. In particular, we focus on `NIFS::prove`, which allocates new vectors to compute the fold step, when technically we could be reusing and overwriting the already allocated vectors from the previous step. Concretely, we add the functions `NIFS::prove_mut` and `R1CSRelaxedWitness/Instance::prove_mut`. We also add a `ResourceSink` with preallocated buffers for `ABC_Z{1,2}` and `T`s to help as stratch memory for the `commit_T_into` operation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,12 +29,15 @@ pub mod supernova;
 
 use once_cell::sync::OnceCell;
 
-use crate::bellpepper::{
-  r1cs::{NovaShape, NovaWitness},
-  shape_cs::ShapeCS,
-  solver::SatisfyingAssignment,
-};
 use crate::digest::{DigestComputer, SimpleDigestible};
+use crate::{
+  bellpepper::{
+    r1cs::{NovaShape, NovaWitness},
+    shape_cs::ShapeCS,
+    solver::SatisfyingAssignment,
+  },
+  r1cs::R1CSResult,
+};
 use abomonation::Abomonation;
 use abomonation_derive::Abomonation;
 use bellpepper_core::ConstraintSystem;
@@ -268,6 +271,20 @@ where
   }
 }
 
+/// A resource sink for [`RecursiveSNARK`]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(bound = "")]
+pub struct ResourceSink<E: Engine> {
+  l_w: Option<R1CSWitness<E>>,
+  l_u: Option<R1CSInstance<E>>,
+
+  ABC_Z_1: R1CSResult<E>,
+  ABC_Z_2: R1CSResult<E>,
+
+  /// buffer for `commit_T`
+  T: Vec<E::Scalar>,
+}
+
 /// A SNARK that proves the correct execution of an incremental computation
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(bound = "")]
@@ -286,6 +303,10 @@ where
   r_U_secondary: RelaxedR1CSInstance<E2>,
   l_w_secondary: R1CSWitness<E2>,
   l_u_secondary: R1CSInstance<E2>,
+
+  sink_primary: ResourceSink<E1>,
+  sink_secondary: ResourceSink<E2>,
+
   i: usize,
   zi_primary: Vec<E1::Scalar>,
   zi_secondary: Vec<E2::Scalar>,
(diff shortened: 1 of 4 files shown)


src/r1cs/mod.rs
```diff
@@ -339,6 +373,56 @@ impl<E: Engine> R1CSShape<E> {
...
+    // this doesn't allocate memory but has bad temporal cache locality -- should test to see which is faster
+    T.clear();
+    tracing::info_span!("T").in_scope(|| {
+      (0..AZ_1.len())
+        .into_par_iter()
+        .map(|i| {
+          let AZ_1_circ_BZ_2 = AZ_1[i] * BZ_2[i];
+          let AZ_2_circ_BZ_1 = AZ_2[i] * BZ_1[i];
+          let u_1_cdot_CZ_2 = U1.u * CZ_2[i];
+          AZ_1_circ_BZ_2 + AZ_2_circ_BZ_1 - u_1_cdot_CZ_2 - CZ_1[i]
```
adr1anh, line 418: Very small nit: subtraction is ever so slightly more expensive than addition. Computing 
`u_1_cdot_Cz_2_plus_Cz_1 = U1.u * CZ_2[I] + CZ_1[i]` and returning 
`AZ_1_circ_BZ_2 + AZ_2_circ_BZ_1 - u_1_cdot_Cz_2_plus_Cz_1` would replace one subtraction by an addition.

adr1anh (commented):
These allocations were really nagging me, I'm glad you're taking care of this! The changes look fine to me, and I think this is a good atomic change to get in soon.

src/lib.rs
```diff
@@ -430,16 +471,25 @@ where
       return Ok(());
     }
 
+    // save the inputs before proceeding to the `i+1`th step
+    let r_U_primary_i = self.r_U_primary.clone();
+    let r_U_secondary_i = self.r_U_secondary.clone();
+    // let l_u_primary_i = self.l_u_primary.clone();
```
huitseeker, line 477: Is this intentional?

src/lib.rs
```diff
@@ -268,6 +271,20 @@ where
   }
 }
 
+/// A resource sink for [`RecursiveSNARK`]
```
huitseeker, line 274: I don't think "resource sink" is known vocabulary you can expect the developer to know. 

In fact, while the elements contained here are the destination of a particular iterations of computation of a `foo_into` function, they are not the final destination of the computed data, and so I'm not sure if the word [*sink*](https://en.wikipedia.org/wiki/Sink_(computing)) is appropriate.

I'd like to see a short comment stating what this is used for without pre-requisites on the part of the reader: storing scratch values computed in a tight loop, which allows reusing the corresponding memory allocations.

src/lib.rs
```diff
@@ -286,6 +303,10 @@ where
   r_U_secondary: RelaxedR1CSInstance<E2>,
   l_w_secondary: R1CSWitness<E2>,
   l_u_secondary: R1CSInstance<E2>,
+
```
huitseeker, line 306: A short comment here recalling what those are for would help.

huitseeker (commented):
I think this is a good change. I would also like to see a PR showing a slightly more extensive variant of that where we delete (instead of adapting, as you do here with `multiply_witness_unchecked`) the allocating methods that this would make obsolete.

If we're going to go for fine-grained allocation control, let's do it fully and not maintain "shadow" allocating methods that would be left unused.

winston-h-zhang:
@huitseeker The allocating methods are still required by SuperNova, and I wanted to avoid a spill-over into changing SuperNova. My plan was to wait once this is merged in, then open the same set of PRs for SuperNova, and everything should be cleaned up.

src/r1cs/mod.rs
winston-h-zhang: Will do!

src/lib.rs
winston-h-zhang: I renamed to `ResourceBuffer`, which should be more appropriate

src/lib.rs
huitseeker: @winston-h-zhang ping

winston-h-zhang: Oops, one second

huitseeker (approved):
🎉

PR #128, opened 2023-11-21
feat: multilinear polynomial evaluation test

huitseeker:
- Expanded the test suite to include test cases asserting the equivalence of MLP evaluation and partial evaluation functions
- Implemented partial evaluation functionality for MLPs in the `multilinear.rs` file
- Enhanced the clarity of the MLP evaluation process by testing it on partial points.

Test tooling:
- Introduced new functions for generating and evaluating Multilinear Polynomials (MLPs) in the `multilinear.rs` file
- Incorporated the use of `ChaCha20Rng`, `SeedableRng`, `RngCore`, and `CryptoRng` libraries for random number generation pertaining to MLP evaluation

PR #147, opened 2023-11-28
Documentation and Naming

porcuquine:
In the course of coming up to speed on things, I added some notes, documentation, tests, and renaming. This was mostly for my own benefit, but it has remained useful — and I think some version of this will be helpful for others approaching from scratch. @adr1anh and @mpenciak please feel free to push commits that fix anything I've said wrong here; or removing anything you think definitely doesn't belong. That said, everything here was useful for me, and I wrote it down for that reason.


src/spartan/polys/multilinear.rs
```diff
@@ -78,6 +85,15 @@ impl<Scalar: PrimeField> MultilinearPolynomial<Scalar> {
...
       .for_each(|(a, b)| {
+        // The right evals include all terms 'active' in the non-top vars (left evals) plus the corresponding terms of the top var.
+        // To isolate only the relevant terms of the top var, we subtract out `a` to get `b - a`.
+        // Then we scale these top-var terms by r: r * (b - a).
+        // Finally we add back in the low terms.
+        // Put differently, since we only want to bind the top var, we subtract the other terms before scaling it, and add them back afterward.
+        // This gives a + r(b - a) as the general expression.
+        // We iterate over all low terms, adjusting them for the bound top var, whose evals will be removed.
+        // This has the net effect of leaving the sum of the new evals equal to the sum of the old evals when the top var is bound to `r` -- as desired.
+        // [This algorithm is described in the proof of Lemma 4.3 of Thaler's Proofs, Arguments, and Zero-Knowledge.]
```
huitseeker, line 96: A general remark is that these comments go into a lot of detail, as they assume little prior knowledge from the reader. This has good and bad sides.

One alternative way of making the code more approachable may be to lean on documents where the algorithmics are explained in more detail, as is done here. Should we want to take this approach, Lemma 3.8 of PAZK comes to mind for the memoized multilinear extension, and I have also found this note helpful: http://bgillespie.com/doc/multilinear-evaluation.pdf

src/spartan/polys/multilinear.rs
```diff
@@ -325,4 +341,36 @@ mod tests {
...
+      let bits = to_bits_be(s, i);
+
+      let e = p.evaluate(&bits);
+
+      assert_eq!(&e, eval);
+    }
+  }
+
+  #[test]
+  fn test_layout() {
```
huitseeker, line 373: https://github.com/lurk-lab/arecibo/pull/128 was meant to make this explicit as well.

src/spartan/polys/eq.rs
```diff
@@ -55,11 +63,23 @@ impl<Scalar: PrimeField> EqPolynomial<Scalar> {
...
         .zip(evals_right.par_iter_mut())
-        .for_each(|(x, y)| {
-          *y = *x * r;
-          *x -= &*y;
+        .for_each(|(running_product, term)| {
+          // Most-significant bits' evals come first. Each new eval is a product of the previous, and this bit's term --
+          // which is the evaluation, at 0 or 1, of (e_i * x_i + (1 - e_i) * (1 - x_i)).
+          // Substituting r for e_i: (r * x_i + (1 - r) * (1 - x_i))
+          // x_i is 'this bit', which is either 0 or 1.
+          //
```
huitseeker, line 71: This might refer more directly to Lagrange basis polynomials (PAZK circa Lemma 3.6)

src/spartan/polys/eq.rs
```diff
@@ -75,16 +95,20 @@ mod tests {
   use pasta_curves::Fp;
 
   fn test_eq_polynomial_with<F: PrimeField>() {
-    let eq_poly = EqPolynomial::<F>::new(vec![F::ONE, F::ZERO, F::ONE]);
+    // Most-significant 'bit' first.
+    let eq_poly = EqPolynomial::<F>::new(vec![F::ONE, F::ZERO, F::ZERO]);
```
huitseeker, line 99: Yes, good change: testing this with palindromic MLEs is pointless and already bit me once.

src/spartan/polys/eq.rs
mpenciak: I agree with Francois that it might be worth replacing this explanation with a reference to the calculation using Lagrange polynomials (as explained in [`polys/multilinear.rs`](https://github.com/lurk-lab/arecibo/blob/4aa21762a112d47d44d2bda8017e221a38b67cf0/src/spartan/polys/multilinear.rs#L29))

src/spartan/sumcheck.rs
```diff
@@ -105,21 +105,43 @@ impl<G: Group> SumcheckProof<G> {
...
+                //
+                // The target is the evaluation that will be used to create a univariate-polynomial whose evaluation at
+                // the target is the sum of the combined non-top evaluations.
+                //
+                // The following `reduce `operation effects the sum.
+                //
+                // The required targets are the evaluation points 0 and 2.
+                //
+                // Evaluation at 0 is unaffected by the top var (to be bound).
+                // eval 0: bound_func is A(low)
```
mpenciak: This comment is confusing. I believe I understand what you're getting at, but I think an explanation like this could be clearer:

> This calculates the univariate polynomial obtained by summing over all but the top var.
> The univariate polynomial is specified by its evaluations at 0, and 2 which are obtained by combining the evaluations of each multivariate polynomial at 0 and 2 with `comb_func`. 

It also may be worth mentioning why only the evaluations of 0 and 2 are needed (quadratic polynomial + 1 can be derived from the hint)

src/spartan/polys/eq.rs
porcuquine: I'm open to reworking, but I'm not sure why that's most relevant here. I'm just inlining the Eq polynomial evaluation formula as given in this module's doc. Please feel free to just write something that improves on this, if it's easy for you. The above is just meant to be a literal unpacking of the code to understand how it does what it claims. [NOTE: I'm disclaiming the background knowledge to write an optimally pedantic explanation and inviting anyone who possesses it do so. What I wrote explains what was missing for *me*.]

porcuquine:
At this point, the global optimum is probably just to leave this as an individual rite of passage for future adventurers.

PR #148, opened 2023-11-28
Only create taus once.

porcuquine:
Don't create all the powers for every `PowPolynomial` when batching. Instead create all the powers once and copy the ones needed. Also, don't even make the `PowPolynomial` (cloning all the powers), when only doing so to get the evals.

This may not matter much, since I guess the concrete number of powers grows logarithmically so may be smallish. Nevertheless, if we aspire to very large circuit sets, it may be worth thinking this way.

### DIFF
--- a/src/spartan/polys/power.rs
+++ b/src/spartan/polys/power.rs
@@ -2,6 +2,7 @@
 
 use crate::spartan::polys::eq::EqPolynomial;
 use ff::PrimeField;
+use std::iter::successors;
 
 /// Represents the multilinear extension polynomial (MLE) of the equality polynomial $pow(x,t)$, denoted as $\tilde{pow}(x, t)$.
 ///
@@ -10,7 +11,6 @@ use ff::PrimeField;
 /// \tilde{power}(x, t) = \prod_{i=1}^m(1 + (t^{2^i} - 1) * x_i)
 /// $$
 pub struct PowPolynomial<Scalar: PrimeField> {
-  t_pow: Vec<Scalar>,
   eq: EqPolynomial<Scalar>,
 }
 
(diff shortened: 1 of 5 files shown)


src/spartan/polys/power.rs
```diff
@@ -30,6 +30,24 @@ impl<Scalar: PrimeField> PowPolynomial<Scalar> {
...
+  /// Create powers the following powers of `t`:
+  /// [t^{2^0}, t^{2^1}, ..., t^{2^{ell-1}}]
+  pub(crate) fn powers(t: &Scalar, ell: usize) -> Vec<Scalar> {
+    let mut t_pow = vec![Scalar::ONE; ell];
+    t_pow[0] = *t;
+    for i in 1..ell {
+      t_pow[i] = t_pow[i - 1].square();
+    }
+    t_pow
+  }
```
huitseeker, line 42: Aka `std::iter::successors(Scalar::ONE, |p| p.square()).take(ell).collect::<Vec<_>>()`
https://doc.rust-lang.org/std/iter/fn.successors.html

src/spartan/polys/power.rs
```diff
@@ -30,6 +30,24 @@ impl<Scalar: PrimeField> PowPolynomial<Scalar> {
...
+
+  /// Creates the evals corresponding to a `PowPolynomial` from an already-existing vector of powers.
+  /// `t_pow.len() > ell` must be true.
+  pub(crate) fn evals_with_powers(powers: &[Scalar], ell: usize) -> Vec<Scalar> {
+    let t_pow = powers[..ell].to_vec();
```
huitseeker, line 47: This is a clone: not sure if this is usable here, but https://docs.rs/ref-cast/latest/ref_cast/ allows (safe-)casting a `&[Scalar]` to a `&EqPolynomial` if it is enough to get us by.

huitseeker (approved):

src/spartan/polys/power.rs
porcuquine: It doesn't quite work in practice. I did some other juggling to remove the remaining clone, I think, in the latest.

PR #151, opened 2023-11-30
SuperNova: Check W size

adr1anh:
This PR adds another Sumcheck instance for checking that W has the correct length.


huitseeker (changes requested):
Please rebase on the latest batched_spartan (this should be easy)

huitseeker (approved):
🖖

PR #157, opened 2023-12-01
Avoid EqPolynomial when only for evals.

porcuquine:
This eliminates some unnecessary cloning caused by unused intermediate `EqPolynomial`s.


porcuquine:
I'll look into reworking with `ref_cast` but merging this now to simplify dev for all.

PR #158, opened 2023-12-01
More zip with

porcuquine:
This adds a more general form of `zip_with` to allow specifying the iterator-creation method and the post-zip function — then creates and uses the most common cases. This gets closer to the goal of making source legible by factoring out noise associated with zipping transformation.

This PR incidentally includes #157.


huitseeker (commented):
## TL;DR

I think if we want to be really extensive with the `zip_with` macro, we should do it with a separate crate solving the general problem, which is fundamentally a rust problem. The reason I wrote the `zip_with` macro was because it was a few lines' worth of change: if we want to be extensive about this, a proc_macro would allow solving all the below issues.

## In Detail
- I think the main lack of expressivity in the macro lies in the rewriting of the argument-pattern in the closure, which does not allow anything but an [identifier pattern](https://doc.rust-lang.org/reference/patterns.html),
- it's unclear to me that we want to systematically use the same iterator for each of the zipped collections, or more exactly the fact we do now may be a consequence of the above limitation: it often makes sense to operate mutably on the first argument, and by reference on all others,
- in all rigor, this should also allow changing the combinator that's used to apply the closure afterwards. Since you added `for_each` previously, the main omission seems to be `flat_map` at this stage.

porcuquine:
I think you're right that it would be nice to factor this out into its own crate. I think it's also true that a proc macro is what it will take to solve this optimally and reach the 'one macro to rule them all'. It's *not* entirely clear what the desired syntax of that macro should be to leave the resulting uses concise, clear, legible, etc.

I think the present PR does a reasonable job of factoring out the common cases into a usable surface syntax. The main incremental improvement I would hope to see in a super-general macro would be support for non-uniform 'iterator-producers', and potentially support for sequential zip -> combine -> zip -> combine operations, as we see in some places.

I don't think the frequent use of the same iterator type is an artifact of the macro(s). Rather, the macros we have now are those that handle the bulk of cases in the code as written before the macros existed.

I'm not sure what you're saying about flat_map: this PR includes a usage of both `zip_with_flat_map!` (https://github.com/lurk-lab/arecibo/pull/158/files#diff-7933bd5a2230aaedf8fb8e64bfdf0055883e63b9ee1ce8ff59921f45eea0ce60R728) and of `zip_with_fn!` where `[flatmap]` is specified (https://github.com/lurk-lab/arecibo/pull/158/files#diff-7933bd5a2230aaedf8fb8e64bfdf0055883e63b9ee1ce8ff59921f45eea0ce60R698-R707).

I'd suggest that the pragmatic path through this is:

- merge this PR
- refactor to a separate crate within arecibo in a separate PR
- over time, refactor toward some notional perfect-macro implementation
- expatriate the perfected macro to its own repo and release, using the result here.

huitseeker (approved):
OK, let's see where this goes and iterate. The only thing I'd insist on is the macros need to be in their own documented  / tested module, even if this doesn't have to be done in this PR.

huitseeker:
See also #164

mpenciak (approved):

PR #161, opened 2023-12-05
Adhere to `C-Caller-Control`

mpenciak:
This PR tries to address the comments about [C-Caller-Control](https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control) API design.

I noticed some further changes that could be made, but they would require changes in code unrelated to Supernova/CompressedSNARKs. As such I decided not to touch them for now.

### DIFF
--- a/benches/compressed-snark-supernova.rs
+++ b/benches/compressed-snark-supernova.rs
@@ -173,7 +173,7 @@ fn bench_one_augmented_circuit_compressed_snark(c: &mut Criterion) {
     // Benchmark the verification time
     group.bench_function("Verify", |b| {
       b.iter(|| {
-        assert!(black_box(&compressed_snark)
+        assert!(black_box(compressed_snark.clone())
           .verify(
             black_box(&pp),
             black_box(&verifier_key),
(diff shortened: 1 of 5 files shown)


huitseeker (commented):
So :
1. the guideline being applied here is: we should not have methods that take their arguments by reference and will systematically make copies of their arguments, because they take away control of how those copies are made from the caller (C-CALLER-CONTROL). Instead if we write methods that need allocated arguments, they should take these arguments by value.
2. here we change the trait implementations in `BatchedRelaxedR1CSSNARKTrait` to take some of their arguments by value whereas they took those arguments by reference before this PR -- in application of (1.)
3. so presumably there are at least *some* implementations of `BatchedRelaxedR1CSSNARKTrait::prove` and `BatchedRelaxedR1CSSNARKTrait::verify` that were doing systematic clones of the `S, U, V` arguments, and no longer need to: they're being passed values. 
4. Why can't I see those clones disappear in the present PR?
5. If there are no clones of these arguments being made in *any* of the implementations of `{prove, verify}`, there is no need to take those arguments by value, is there?

huitseeker (changes requested):
I've had a look at this, and due essentially to the (over-?)use of the field's instance of `Copy`, the copies that can be eliminated are few (in proportion to all copies).

I suggest we table this upheaval of the fundamental `RelaxedR1CSNARKTrait` until later, at a time where we'll be able to form a cogent strategy to unify the trait API upstream.

PR #162, opened 2023-12-05
Introduce proper `MaskedEq` poly

adr1anh:
This PR introduces a new `MaskesEqPolynomial` which is equal to an `EqPolynomial` over the same $n$ random values, except over the first $2^m$ evaluations where it equals zero. 

We use it in the `WitnessBoundSumcheck` to ensure the `W` polynomial only contains values in the first `num_vars` entries.

### DIFF
--- a/src/spartan/polys/masked_eq.rs
+++ b/src/spartan/polys/masked_eq.rs
@@ -1,0 +1,149 @@
+//! `MaskedEqPolynomial`: Represents the `eq` polynomial over n variables, where the first 2^m entries are 0.
+
+use crate::spartan::polys::eq::EqPolynomial;
+use ff::PrimeField;
+use itertools::zip_eq;
+
+/// Represents the multilinear extension polynomial (MLE) of the equality polynomial $eqₘ(x,r)$
+/// over n variables, where the first 2^m evaluations are 0.
+///
+/// The polynomial is defined by the formula:
+/// eqₘ(x,r) = eq(x,r) - ( ∏_{0 ≤ i < n-m} (1−rᵢ)(1−xᵢ) )⋅( ∏_{n-m ≤ i < n} (1−rᵢ)(1−xᵢ) + rᵢ⋅xᵢ )
+pub struct MaskedEqPolynomial<Scalar: PrimeField> {
+  eq: EqPolynomial<Scalar>,
+  num_masked_vars: usize,
+}
+
+impl<Scalar: PrimeField> MaskedEqPolynomial<Scalar> {
+  /// Creates a new `MaskedEqPolynomial` from a vector of Scalars `r` of size n, with the number of
+  /// masked variables m = `num_masked_vars`.
+  pub const fn new(r: Vec<Scalar>, num_masked_vars: usize) -> Self {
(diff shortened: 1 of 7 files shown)


src/spartan/polys/masked_eq.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Represents the multilinear extension polynomial (MLE) of the equality polynomial $eqₘ(x,r)$
+/// over n variables, where the first 2^m evaluations are 0.
+///
+/// The polynomial is defined by the formula:
+/// eqₘ(x,r) = eq(x,r) - ( ∏_{0 ≤ i < n-m} (1−rᵢ)(1−xᵢ) )⋅( ∏_{n-m ≤ i < n} (1−rᵢ)(1−xᵢ) + rᵢ⋅xᵢ )
+pub struct MaskedEqPolynomial<Scalar: PrimeField> {
+  eq: EqPolynomial<Scalar>,
+  num_masked_vars: usize,
+}
```
huitseeker, line 15: Should this be?
```rust
pub struct MaskedEqPolynomial<'a, Scalar: PrimeField> {
  eq: &'a EqPolynomial<Scalar>,
  num_masked_vars: usize,
}
```

src/spartan/polys/masked_eq.rs
```diff
@@ -0,0 +1,149 @@
...
+use crate::spartan::polys::eq::EqPolynomial;
+use ff::PrimeField;
+use itertools::zip_eq;
+
+/// Represents the multilinear extension polynomial (MLE) of the equality polynomial $eqₘ(x,r)$
+/// over n variables, where the first 2^m evaluations are 0.
+///
+/// The polynomial is defined by the formula:
+/// eqₘ(x,r) = eq(x,r) - ( ∏_{0 ≤ i < n-m} (1−rᵢ)(1−xᵢ) )⋅( ∏_{n-m ≤ i < n} (1−rᵢ)(1−xᵢ) + rᵢ⋅xᵢ )
+pub struct MaskedEqPolynomial<Scalar: PrimeField> {
```
huitseeker, line 12: This needs to derive `Debug` at least.

huitseeker (commented):
1. Do we want to manipulate an owned `EqPolynomial` or a reference to such an `EqPolynomial`? What I'm getting at is that if the `MaskedPolynomial` only exists as a thin wrapper on the way to getting its `.evals()`, https://github.com/lurk-lab/arecibo/pull/157 has shown (on the `EqPolynomial`) that this pattern leads to uneeded copies.
2. If the answer is "yeah, there's an upcoming instance where I will need to act on this `MaskedPolynomial` mutably", have you considered a cow ? (as in [Cow](https://doc.rust-lang.org/std/borrow/enum.Cow.html) not 🐄 )

mpenciak (approved):

huitseeker (changes requested):
I had a quick look and I can confirm that `MaskedEqPolynomial` operating by reference does avoid needless cloning. I've opened https://github.com/adr1anh/arecibo/pull/2

huitseeker (approved):
🙏

PR #165, opened 2023-12-05
ci: Add trigger for PRs to batched_spartan

samuelburnham:


huitseeker (changes requested):
~~The job that runs is the one that's defined on the main branch of the repo, so those changes would be ineffectual, you need to rebase on dev.~~
Edit: that's not true for `pull_request` events. TIL

huitseeker (approved):

PR #182, opened 2023-12-14
apply bound witness fix to ppsnark

adr1anh:


huitseeker (approved):
thank you so much for this fix!

PR #187, opened 2023-12-15
GPU accelerated Grumpkin MSM

### LINKED ISSUE #186: Integrate with grumpkin-msm
https://github.com/lurk-lab/grumpkin-msm should be integrated in https://github.com/lurk-lab/arecibo/blob/dev/src/provider/bn256_grumpkin.rs following the usual pasta_curves pattern

winston-h-zhang:
Resolves #186.

Note: `force-no-sort` does not disable sorting. It removes redundant sort-related symbols between `pasta-msm` and `grumpkin-msm` so the linker is happy.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,8 @@ rand = "0.8.5"
 ref-cast = "1.0.20"
 
 [target.'cfg(any(target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
-pasta-msm = { git="https://github.com/lurk-lab/pasta-msm", branch="dev", version = "0.1.4" }
+pasta-msm = { git = "https://github.com/lurk-lab/pasta-msm", branch = "dev", version = "0.1.4" }
+grumpkin-msm = { git = "https://github.com/lurk-lab/grumpkin-msm", branch = "dev", version = "0.1.0" }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 halo2curves = { version = "0.5.0", features = ["bits", "derive_serde"] }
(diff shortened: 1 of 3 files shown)


huitseeker (commented):
This is good overall, but we would *really* benefit from a test that, on the right platforms (x86/aarch64) and under the right feature (cuda), compares the results of `cpu_best_msm` and `grumpkin::{bn256, grumpkin}` for a deterministically-seeded random point/scalar vector.

This test should probably have existed in the pasta variant in the first place, and will give us a lot more confidence both in the sppark GPU implementation and in CPU changes required by #193

huitseeker (changes requested):
Back to your queue for the missing test.

Cargo.toml
```diff
@@ -47,7 +47,8 @@ rand = "0.8.5"
 ref-cast = "1.0.20"
 
 [target.'cfg(any(target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
-pasta-msm = { git="https://github.com/lurk-lab/pasta-msm", branch="dev", version = "0.1.4" }
+pasta-msm = { git = "https://github.com/lurk-lab/pasta-msm", branch = "dev", version = "0.1.4" }
+grumpkin-msm = { git = "https://github.com/lurk-lab/grumpkin-msm", branch = "dev", features = ["dont-implement-sort"] }
```
huitseeker, line 51: The following comment would elucidate the feature:
```suggestion
# pasta-msm also calls into sppark, which defines the same common sorting code this crate would: the `dont-implement-sort` feature avoid creating conflicting symbols.
grumpkin-msm = { git = "https://github.com/lurk-lab/grumpkin-msm", branch = "dev", features = ["dont-implement-sort"] }
```

huitseeker (approved):
Thank you so much!

PR #195, opened 2023-12-19
Use `expect_test` in `test_pp_digest`, `test_recursive_circuit`, and `test_supernova_recursive_circuit`

### LINKED ISSUE #132: Refactor test_pp_digest to use expected_test
test_pp_digest is a pain to update due to its runtime and the changing nature of our hashing & structures:
https://github.com/lurk-lab/arecibo/blob/dev/src/lib.rs#L1019-L1114
We should refactor it to use expected_test, which would make the update much easier:
https://docs.rs/expect-test/latest/expect_test/

mpenciak:
This closes #132. 

There may be other tests that would benefit from this (maybe the ecc op circuit tests for constraint numbers?)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,7 @@ pprof = { version = "0.11" }
 sha2 = "0.10.7"
 tracing-test = "0.2.4"
 proptest = "1.2.0"
+expect-test = "1.4.1"
 
 [[bench]]
 name = "recursive-snark"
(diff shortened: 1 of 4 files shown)


huitseeker:
> There may be other tests that would benefit from this (maybe the ecc op circuit tests for constraint numbers?)

You're 💯 right. Those test have been introduced as part of https://github.com/lurk-lab/arecibo/pull/146

huitseeker (approved):
Great work, thanks a lot!

PR #200, opened 2023-12-20
Sample random R1CS and satisfying instance/witness

### LINKED ISSUE #199: feat: Sample random R1CS instances
For testing purposes of both the SNARKs and the verifiers, it would be useful to have sample random Relaxed R1CS instances with exact sizes. 

The 3 parameters we want to define are 
- `m = num_constraints` the number of rows of $A,B,C$ and the size of $E$
- `n = num_variables` the size of `W`, and $A,B,C$ have `2*n` columns.
- `l = num_outputs` the number of public inputs, not counting 1
- `N = max{|A|+|B|+|C|}` this is the number of non-zero entries in the matrices $A,B,C$

- Sample 3 random sparse matrices $A,B,C$ of size `(m x 2n)` and fill them up with `N` values, which can be balanced between the matrices.
- Sample a random vector $W \in \mathbb{F}^n$, public output $X \in \mathbb{F}^\ell$, and slack $u \in \mathbb{F}$
- Set $Z = [W, u, X]$ and compute the corresponding $E \in \mathbb{F}^m$ vector as $E = AZ \circ BZ - u\cdot Cz$ 

The tuple $(W, u, X, E)$ is a valid RelaxedR1CSWitness for the matrices $A,B,C$.

adr1anh:
Fixes #199.

### DIFF
--- a/src/r1cs/mod.rs
+++ b/src/r1cs/mod.rs
@@ -20,7 +20,6 @@ use abomonation::Abomonation;
 use abomonation_derive::Abomonation;
 use core::cmp::max;
 use ff::{Field, PrimeField};
-use itertools::zip_eq;
 use once_cell::sync::OnceCell;
 use rand_core::{CryptoRng, RngCore};
 
@@ -199,7 +198,7 @@ impl<E: Engine> R1CSShape<E> {
   }
 
   /// Generate a satisfying [RelaxedR1CSWitness] and [RelaxedR1CSInstance] for this [R1CSShape].
-  pub fn random_witness_instance<E: Engine, R: RngCore + CryptoRng>(
+  pub fn random_witness_instance<R: RngCore + CryptoRng>(
     &self,
     commitment_key: &CommitmentKey<E>,
     mut rng: &mut R,
@@ -339,7 +338,7 @@ impl<E: Engine> R1CSShape<E> {
 
     let E = zip_with!(
       (Az.into_par_iter(), Bz.into_par_iter(), Cz.into_par_iter()),
-      |a, b, c| a * b - u * c
+      |a, b, c| a * b - c * u
     )
     .collect::<Vec<E::Scalar>>();
 
@@ -359,10 +358,12 @@ impl<E: Engine> R1CSShape<E> {
 
     // verify if Az * Bz - u*Cz = E
     let E = self.compute_E(&W.W, &U.u, &U.X)?;
-    zip_eq(W.E.par_iter(), E.par_iter())
+    W.E
+      .par_iter()
+      .zip_eq(E.into_par_iter())
       .enumerate()
       .try_for_each(|(i, (we, e))| {
-        if *we != *e {
+        if *we != e {
           // constraint failed, retrieve constraint name
           Err(NovaError::UnSatIndex(i))
         } else {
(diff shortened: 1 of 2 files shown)


src/r1cs/mod.rs
```diff
@@ -159,6 +160,76 @@ impl<E: Engine> R1CSShape<E> {
     })
   }
 
+  /// Generate a random [R1CSShape] with the specified number of constraints, variables, and public inputs/outputs.
+  pub fn random<R: RngCore + CryptoRng>(
```
huitseeker, line 164: Just a thought, if we're going to implement `random` recursively on some structs, could we possibly use `Arbitrary` from [proptest](https://github.com/proptest-rs/proptest)? We'd get randomized testing as a side-effect.

src/r1cs/mod.rs
```diff
@@ -159,6 +160,76 @@ impl<E: Engine> R1CSShape<E> {
...
+      num_vars,
+      num_io,
+      A,
+      B,
+      C,
+      digest: Default::default(),
+    }
+  }
+
+  /// Generate a satisfying [RelaxedR1CSWitness] and [RelaxedR1CSInstance] for this [R1CSShape].
```
huitseeker, line 200: We might want to signal in a comment if the witness satisfies the instance, as this is relevant to the range of tests (positive, negative) one can write with this.

src/r1cs/sparse.rs
```diff
@@ -85,6 +87,30 @@ impl<F: PrimeField> SparseMatrix<F> {
...
+    while indices.len() < num_entries {
+      let row = rng.next_u32() as usize % rows;
+      let col = rng.next_u32() as usize % cols;
+      indices.insert((row, col));
+    }
+
+    let matrix = indices
+      .into_iter()
+      .map(|(row, col)| (row, col, F::random(&mut rng)))
+      .collect::<Vec<_>>();
```
huitseeker, line 109: I think this may (occasionally) produce duplicates, which is worrisome, because I think the `SparseMatrix` logic does not check for duplicates. That later remark is something we should test, BTW.

Summoning @winston-h-zhang to check what I'm saying.

huitseeker (commented):
This looks good overall, but I think we may need a duplicate check on the NZ coefficients produced by `SparseMatrix::random` (see inline)

storojs72 (commented):
I like this idea and I suspect the motivation for it is the last task related to Solidity implementing the IPA on Grumpkin for ultra-small proving statement (to avoid OutOfGas while MSM).

I would even extend this to the following: implementing `random` recursively in order to be able to derive an arbitrary circuits with pre-defined complexity. And it would be nice if such circuits could be usable with the e2e testing (like in `src/lib.rs`). I guess, that would be super-useful for the mentioned Solidity tasks, where some custom building block needs to be implemented and we need actual test vector for debugging.

Probably that is out-of-scope for this PR

src/r1cs/sparse.rs
huitseeker: Edit: my bad, the `BTreeSet` guarantees this will not produce duplicates, and the while loop guarantees that this will produce exactly `num_entries`,  essentially doing rejection sampling. I'm turning blind.

huitseeker (approved):

huitseeker (approved):

src/r1cs/mod.rs
adr1anh: Do you mean this could be interpreted as "the witness is satisfying but the instance may not"? Or should the method name be clearer?

src/r1cs/mod.rs
adr1anh: I'm having trouble finding documentation on how to do it, but I'll create an issue to tackle it later.

Issue #201, opened 2023-12-20
chore: Remove redundant absorbing to the transcript


storojs72:
One of `CompressedSNARK`'s verification steps (affects only [ppspartan.rs](https://github.com/lurk-lab/arecibo/blob/dev/src/spartan/ppsnark.rs)) - checking the satisfiability of the folded instances ([link](https://github.com/lurk-lab/arecibo/blob/dev/src/lib.rs#L947)) - has [following](https://github.com/lurk-lab/arecibo/blob/dev/src/spartan/ppsnark.rs#L1520) redundant absorption of `eval_vec` to the transcript, since it is already happened [here](https://github.com/lurk-lab/arecibo/blob/dev/src/spartan/ppsnark.rs#L1499). Absorbing the same value twice adds no new entropy to the transcript, so second one can be freely removed. 

Also [affects](https://github.com/microsoft/Nova/blob/main/src/spartan/ppsnark.rs#L1367) upstream Nova implementation.

PR #203, opened 2023-12-20
Removing redundant absorption to the transcript (#201)

storojs72:
This PR fixes https://github.com/lurk-lab/arecibo/issues/201.

It would be nice to apply the patch to [upstream](https://github.com/microsoft/Nova/tree/main) Nova implementation as well.


adr1anh (approved):
Looks good to me, good catch !

huitseeker (approved):

Issue #202, opened 2023-12-20
chore: Remove redundant absorptions to transcript in MLKZG implementation (PR#172)


storojs72:
In https://github.com/lurk-lab/arecibo/pull/172 there is an alternative PCS implemented - MLKZG. @adr1anh left couple of comments related to redundant absorptions to the transcript. Absorbing value, that has already been absorbed before, adds no new entropy to the transcript, so such redundant absorptions can be removed.

Links to commets: 
- https://github.com/lurk-lab/arecibo/pull/172#discussion_r1427743241
- https://github.com/lurk-lab/arecibo/pull/172#discussion_r1427819475
- https://github.com/lurk-lab/arecibo/pull/172#discussion_r1427852516

PR #172, opened 2023-12-11
Add MLKZG support (Nova forward port)

huitseeker:
This forward-ports the following Nova PRs:
- https://github.com/microsoft/Nova/pull/269 (MLKZG support)
- https://github.com/microsoft/Nova/pull/272 (minroot on MLKZG)

> [!WARNING]  
> This is not just a typical forward port from Nova. The second commit includes a significant, opinionated refactor of the MLKZG computation traits to reduce redundancy and boilerplate code.

> [!NOTE]
> Items left to address in a future PR include:
> - the `TranscriptReprTrait` bounds that have been left on `DlogGroup` since https://github.com/microsoft/Nova/pull/263 are surprisingly hard to use and should probably be relocated,
> - Refactoring the `UniversalKZGParam` struct and abstracting it behind a trait, given the varying public parameter requirements in different schemes (e.g., MLKZG requires fewer G2 points).
> - ~~Renaming `UVKZGProverKey` and `UVKZGVerifierKey` as they serve dual roles in multiple multilinear schemes.~~
> - ~~Fixing the discrepancy between `TranscriptReprTrait<G> for $name::Affine` and `TranscriptReprTrait<E::GE> for Commitment<E>` and `TranscriptReprTrait<E::G1> for UVKZGCommitment<E>` (only the last 2 agree with each other)~~

## Minroot tests

We get a slight performance improvement *through the refactoring on the second commit*, as seen on the Minroot example:
download (2). The machine is an M1 Mac (CPU only), and the "dataset size" is the number of Minroot iterations in the folded circuit (multiply by roughly 10^3 to get the number of constraints, go see the full log to get exact numbers).

<details><summary>
Minroot example highlights</summary>

```
huitseeker@lightning.localdomain➜~/tmp/nova(port_mlkzg)» curl https://gist.githubusercontent.com/huitseeker/ecd67d7fa6cd1640ed244dc6131fd73f/raw/6bd7d80a8222c9174563599b570b5b316b850fc6/before.log |grep -Ee 'CompressedSNARK.*?(prove|verify)'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  8981  100  8981    0     0   137k      0 --:--:-- --:--:-- --:--:--  143k
CompressedSNARK::prove: true, took 15.17823149s
CompressedSNARK::verify: true, took 853.667266ms
CompressedSNARK::prove: true, took 15.770005823s
[...]

PR #206, opened 2023-12-20
chore: Remove redundant absorptions in MLKZG implementation

storojs72:
This PR attempts to address @adr1anh 's comments from https://github.com/lurk-lab/arecibo/pull/172 , related to transcript absorptions. 

Fixes https://github.com/lurk-lab/arecibo/issues/202

### DIFF
--- a/src/provider/mlkzg.rs
+++ b/src/provider/mlkzg.rs
@@ -53,9 +53,7 @@ where
   E::Fr: TranscriptReprTrait<E::G1>,
   E::G1Affine: TranscriptReprTrait<E::G1>, // TODO: this bound on DlogGroup is really unusable!
 {
-  fn compute_challenge(
-    transcript: &mut impl TranscriptEngineTrait<NE>,
-  ) -> E::Fr {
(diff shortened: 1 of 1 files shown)


src/provider/mlkzg.rs
```diff
@@ -53,29 +53,17 @@ where
...
 {
-  fn compute_challenge(
-    C: &E::G1Affine,
-    y: &E::Fr,
-    com: &[E::G1Affine],
-    transcript: &mut impl TranscriptEngineTrait<NE>,
-  ) -> E::Fr {
-    transcript.absorb(b"C", C);
-    transcript.absorb(b"y", y);
-    transcript.absorb(b"c", &com.to_vec().as_slice());
```
adr1anh, line 64: This `com` does need to be part of the transcript

adr1anh (changes requested):

src/provider/mlkzg.rs
storojs72: Yes, agree. Thanks. It is being computed inside the proving / verifying so it should improve entropy of the transcript.

adr1anh (approved):

PR #207, opened 2023-12-20
Fix `mlkzg::EvaluationArgument` fields

adr1anh:


src/provider/mlkzg.rs
```diff
@@ -32,9 +32,9 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
   deserialize = "E::G1Affine: Deserialize<'de>, E::Fr: Deserialize<'de>"
 ))]
 pub struct EvaluationArgument<E: Engine> {
-  evals_r: Vec<E::G1Affine>,
-  evals_neg_r: Vec<E::G1Affine>,
-  evals_r_squared: Vec<Vec<E::Fr>>,
+  comms: Vec<E::G1Affine>,
+  w: Vec<E::G1Affine>,
+  evals: Vec<Vec<E::Fr>>,
```
huitseeker, line 37: if `evals.len()` is known to be 3, an array may make more sense.

huitseeker (commented):
I completely misunderstood your comment, thanks for fixing 🤦

adr1anh:
Given that @storojs72 is working on it at the moment, I just wanted to get this quick fix in so the code is clearer. I agree that we should probably use an array instead though.

huitseeker (approved):

PR #210, opened 2023-12-21
chore: Avoid absorbing 'v' while 'q' computing

storojs72:
As discussed with @adr1anh , we can avoid absorbing `v` while `q` computing because it is deterministically derived from all the previous sent messages.


storojs72:
This is incorrect, I misunderstood @adr1anh

PR #214, opened 2023-12-21
ci: use reuseable workflows

huitseeker:


.github/workflows/rust.yml
```diff
@@ -67,49 +67,16 @@ jobs:
...
-        run: cargo build --target wasm32-unknown-unknown
-      - name: Check Rustfmt Code Style
-        uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
-      - name: Check clippy warnings
-        run: cargo xclippy -D warnings
-      - name: Check benches compile under the flamegraph feature
-        run: cargo check --profile dev-ci --features "flamegraph" --benches
```
samuelburnham, line 98: This flamegraph check isn't contained in the reusable workflow, so it would need a separate job

samuelburnham (changes requested):

.github/workflows/rust.yml
```diff
@@ -67,49 +69,16 @@ jobs:
       working-directory: ${{ github.workspace }}/lurk
```
samuelburnham, line 69: We can also use the https://github.com/lurk-lab/ci-workflows/blob/main/.github/workflows/check-lurk-compiles.yml workflow here, or wait till it's more versatile as it only accepts a `path` and optional `package` keyword for the patch

samuelburnham (dismissed):

huitseeker:
Pausing work on this: we need to switch to https://github.com/dtolnay/rust-toolchain first, before we use our reusable workflows.

.github/workflows/rust.yml
huitseeker: > wait till it's more versatile as it only accepts a path and optional package keyword for the patch

I'm not sure I understand what you're suggesting?

samuelburnham: Future work could be to refine the `check-lurk-compiles` reusable workflow to account for things like arecibo having a `package` Cargo rename, different features, etc. It's not related to this PR

samuelburnham (approved):

Issue #212, opened 2023-12-21
feat: create test vectors & benchmark for PCS


adr1anh:
A multi-linear polynomial commitment scheme is parametrized by 
- Number of variables $n$
- A polynomial `P` $P(X_1,\ldots,X_n)$ given as `Vec<F>` of size $2^n$
- An evaluation point `x` $\vec{x} = (x_1,\ldots,x_n)$ given as `Vec<F>` of size $n$.
- The commitment `C` $C = \mathsf{Commit}(P)$
- The evaluation `e` $e = P(\vec{x})$

The goal is to support generating sampling these parameters in order to derive test vectors (along with the `EvaluationArgument` proofs) for testing in Solidity, as well as benchmarking of the relative performance of different PCS `EvaluationEngine` (in particular `mlzkg` vs `zeromorph`). 

- Implement support for generating random instances of polynomials and evaluation points, and derive both valid and invalid commitments and evaluations.
- Fix a seed and generate serializable vectors of the above data, and a proof for each PCS engine.
- Write a benchmark comparing `mlkzg::EvaluationEngine`, `non_hiding_zero_morph::ZMPCS` and `ipa_pcs::EvaluationEngine` in terms of 
  - Proof size
  - Prover time
  - Verifier time

tchataigner:
Hey all! Happy new year! I've already produced the [generic test methods](https://github.com/tchataigner/arecibo/blob/feature/test-ipa/src/provider/util/mod.rs#L6-L144) along with the [start of a benchmark](https://github.com/tchataigner/arecibo/blob/feature/test-ipa/benches/pcs.rs). 

Refactoring aside, I have some questions about how I did thing:
- I wanted to create a generic test function to assert that `verify` would fail on an unvalid proof. This resulted in [`test_fail_bad_proof`](https://github.com/tchataigner/arecibo/blob/feature/test-ipa/src/provider/util/mod.rs#L104-L143). However, I'm not really sure that what I did is valid way to do so, either on the actual implementation or on the heaviness of the code. What is your opinion?
- In the benchmarks and my test, I currently iteretate over three seeds : 4, 5 ,6. I re-used them from a previous test in MLKZG. I was wondering if you had more precise directions on how many I should run and of which order for the different PCS?

adr1anh:
Thanks for taking up this issue! Will take a deeper look later, but let me answer your questions first. 

[...]

PR #222, opened 2024-01-02
Test vectors & benchmark for PCS

tchataigner:
## Goal of this PR

The goal of this PR is to implement test vectors & benchmark for PCS.

### Current progress

Implemented two generic test functions that we can use over any `<Engine, EvaluationEngine<Engine>>` for both the correct prove/verify flow and a bad proof verification flow.

### Left TODO

- [x] Part of the issue is to tackle benchmark for proof size, proving time and verification time.

## Related issue

This is part of tackling the issue #212

### DIFF
--- a/benches/pcs.rs
+++ b/benches/pcs.rs
@@ -98,7 +98,7 @@ macro_rules! benchmark_all_engines {
             // Proving group
             let mut proving_group = $criterion.benchmark_group(format!("PCS-Proving {}", num_vars));
             proving_group
-                .sampling_mode(SamplingMode::Flat)
+                .sampling_mode(SamplingMode::Auto)
                 .sample_size(10);
 
             $(
@@ -112,7 +112,7 @@ macro_rules! benchmark_all_engines {
             // Verifying group
             let mut verifying_group = $criterion.benchmark_group(format!("PCS-Verifying {}", num_vars));
             verifying_group
-                .sampling_mode(SamplingMode::Flat)
+                .sampling_mode(SamplingMode::Auto)
                 .sample_size(10);
 
             $(
--- a/benches/pcs.rs
+++ b/benches/pcs.rs
@@ -13,21 +13,34 @@ use halo2curves::bn256::Bn256;
 use rand::rngs::StdRng;
 use rand_core::{CryptoRng, RngCore, SeedableRng};
 use std::time::Duration;
+
 // To run these benchmarks, first download `criterion` with `cargo install cargo-criterion`.
-// Then `cargo criterion --bench pcs`. The results are located in `target/criterion/data/<name-of-benchmark>`.
-criterion_group! {
-      name = pcs;
-      config = Criterion::default().warm_up_time(Duration::from_millis(3000));
-      targets = bench_pcs
+// Then `cargo criterion --bench pcs`.
+// For flamegraphs, run `cargo criterion --bench pcs --features flamegraph -- --profile-time <secs>`.
+// The results are located in `target/criterion/profile/<name-of-benchmark>`.
+cfg_if::cfg_if! {
+  if #[cfg(feature = "flamegraph")] {
+    criterion_group! {
+          name = pcs;
+          config = Criterion::default().warm_up_time(Duration::from_millis(3000)).with_profiler(pprof::criterion::PProfProfiler::new(100, pprof::criterion::Output::Flamegraph(None)));
+          targets = bench_pcs
+    }
+  } else {
+    criterion_group! {
+          name = pcs;
+          config = Criterion::default().warm_up_time(Duration::from_millis(3000));
+          targets = bench_pcs
+    }
+  }
 }
 
 criterion_main!(pcs);
 
-const TEST_ELL: [usize; 3] = [4, 5, 6];
+const TEST_ELL: [usize; 11] = [10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20];
 fn bench_pcs(c: &mut Criterion) {
   for ell in TEST_ELL.iter() {
-    let mut group = c.benchmark_group(format!("PCS-Seed-{ell}"));
-    group.sampling_mode(SamplingMode::Flat);
+    let mut group = c.benchmark_group(format!("PCS-PolynomialSize-{ell}"));
+    group.sampling_mode(SamplingMode::Flat).sample_size(10);
 
     let mut rng = rand::rngs::StdRng::seed_from_u64(*ell as u64);
 
--- a/benches/pcs.rs
+++ b/benches/pcs.rs
@@ -35,7 +35,7 @@ cfg_if::cfg_if! {
 
 criterion_main!(pcs);
 
-const TEST_ELL: [usize; 2] = [10, 11];
+const TEST_ELL: [usize; 11] = [10, 11, 12, 23, 14, 15, 16, 17, 18, 19, 20];
 
 struct BenchAssests<E: Engine, EE: EvaluationEngineTrait<E>> {
   poly: MultilinearPolynomial<<E as Engine>::Scalar>,
--- a/benches/pcs.rs
+++ b/benches/pcs.rs
@@ -1,0 +1,112 @@
+use arecibo::provider::ipa_pc::EvaluationEngine as IPAEvaluationEngine;
+use arecibo::provider::mlkzg::EvaluationEngine as MLEvaluationEngine;
+use arecibo::provider::non_hiding_zeromorph::ZMPCS;
+use arecibo::provider::util::test_utils::{commitment_setup, gen_poly_point_eval};
+use arecibo::provider::{Bn256EngineKZG, Bn256EngineZM, GrumpkinEngine};
+use arecibo::traits::evaluation::EvaluationEngineTrait;
+use arecibo::traits::{Engine, TranscriptEngineTrait};
+use criterion::measurement::WallTime;
+use criterion::{criterion_group, criterion_main, BenchmarkGroup, Criterion, SamplingMode};
+use halo2curves::bn256::Bn256;
+use std::time::Duration;
+// To run these benchmarks, first download `criterion` with `cargo install cargo-criterion`.
+// Then `cargo criterion --bench pcs --features bench`. The results are located in `target/criterion/data/<name-of-benchmark>`.
+criterion_group! {
+      name = pcs;
+      config = Criterion::default().warm_up_time(Duration::from_millis(3000));
+      targets = bench_proving
+}
+
+criterion_main!(pcs);
+
+const SEEDS: [usize; 3] = [4, 5, 6];
+fn bench_proving(c: &mut Criterion) {
+  for seed in SEEDS.iter() {
+    let mut group = c.benchmark_group(format!("PCS-Seed-{seed}"));
+    group.sampling_mode(SamplingMode::Flat);
+
+    bench_proving_internal::<GrumpkinEngine, IPAEvaluationEngine<GrumpkinEngine>>(
+      &mut group,
+      "IPA-Grumpkin",
+      *seed,
+    );
+
+    bench_proving_internal::<Bn256EngineKZG, MLEvaluationEngine<Bn256, Bn256EngineKZG>>(
+      &mut group, "MLKZG-Bn", *seed,
+    );
+
+    bench_proving_internal::<Bn256EngineZM, ZMPCS<Bn256, Bn256EngineZM>>(
+      &mut group, "ZM-Bn", *seed,
+    );
+
+    group.finish();
+  }
+}
+
+fn bench_proving_internal<E: Engine, EE: EvaluationEngineTrait<E>>(
+  group: &mut BenchmarkGroup<'_, WallTime>,
+  id: &str,
+  seed: usize,
+) {
+  let (poly, point, eval) = gen_poly_point_eval::<E>(seed);
+
+  let (ck, commitment, prover_key, verifier_key) = commitment_setup::<E, EE>(seed, &poly);
+
+  // Generate proof.
+  group.bench_function(format!("Prove-{}", id), |b| {
+    b.iter(|| {
+      assert!(EE::prove(
+        &ck,
+        &prover_key,
+        &mut E::TE::new(b"TestEval"),
+        &commitment,
+        poly.evaluations(),
+        &point,
+        &eval,
+      )
+      .is_ok());
+    })
+  });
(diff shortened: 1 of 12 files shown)


src/provider/non_hiding_zeromorph.rs
```diff
@@ -514,87 +514,30 @@ where
...
-      traits::DlogGroup,
-      Bn256Engine,
+      non_hiding_kzg::UVKZGPoly,
+      non_hiding_zeromorph::{batched_lifted_degree_quotient, eval_and_quotient_scalars, ZMPCS},
     },
     spartan::polys::multilinear::MultilinearPolynomial,
-    traits::{Engine as NovaEngine, Group, TranscriptEngineTrait, TranscriptReprTrait},
   };
 
-  fn commit_open_verify_with<E: MultiMillerLoop, NE: NovaEngine<GE = E::G1, Scalar = E::Fr>>()
```
huitseeker, line 543: this test is quite useful and should be preserved, as a test, in its generic form (see the associated `test_commit_open_verify` which constitutes a formal rust test for this generic function)

tchataigner: Sorry, I misinterpreted what it did and thought it only tested proving and verifying. Adding it back;

tchataigner:
I think that I've got the bulk of the logic down. 

A sample of what the benches look like can be found here: 

Now, I would like to iterate on what we actually display on the benchmark and how it should be formatted. Currently, the results are divided in two groups: proving and verifying. They are displayed per the power of two used to generate the polynomial we are testing over.

I'm not sure this fully corresponds to what we wanted. I started to look into the [Throughput measurements](https://bheisler.github.io/criterion.rs/book/user_guide/advanced_configuration.html#throughput-measurements) to capture proof size but I'm not sure on how to capture a proof size in a generic manner.

benches/pcs.rs
```diff
@@ -0,0 +1,181 @@
...
+  ck: <<E as Engine>::CE as CommitmentEngineTrait<E>>::CommitmentKey,
+  commitment: <<E as Engine>::CE as CommitmentEngineTrait<E>>::Commitment,
+  prover_key: <EE as EvaluationEngineTrait<E>>::ProverKey,
+  verifier_key: <EE as EvaluationEngineTrait<E>>::VerifierKey,
+  proof: Option<<EE as EvaluationEngineTrait<E>>::EvaluationArgument>,
+}
+
+// Macro to generate benchmark code for multiple engine and evaluation engine types
+macro_rules! benchmark_engines {
+    ($ell:expr, $rng:expr, $group:expr, $internal_fn:expr, $( ($engine:ty, $eval_engine:ty, $engine_name:expr) ),*) => {
```
huitseeker, line 53: You can use [`type_name!`](https://doc.rust-lang.org/std/any/fn.type_name.html) to avoid the need for the last parameter

benches/pcs.rs
```diff
@@ -0,0 +1,181 @@
...
+        $(
+            let mut assets = deterministic_assets::<$engine, $eval_engine, StdRng>($ell, &mut $rng);
```
huitseeker, line 55: Once you populate the `eval_engine`, type inference should populate the first parameter. IOW, this should work and remove the need for the `engine` param:
`deterministic_assets::<_, $eval_engine, StdRng>($ell, &mut $rng);`

src/provider/util/mod.rs
```diff
@@ -1,3 +1,97 @@
...
+pub mod test_utils {
+  //! Contains utilities for testing and benchmarking.
+  use crate::spartan::polys::multilinear::MultilinearPolynomial;
+  use crate::traits::{
+    commitment::CommitmentEngineTrait, evaluation::EvaluationEngineTrait, Engine,
+  };
+
+  /// Generates a random polynomial and point from a seed to test a proving/verifying flow of one
+  /// of our EvaluationEngine over a given Engine.
+  pub(crate) fn prove_verify_from_ell<E: Engine, EE: EvaluationEngineTrait<E>>(ell: usize) {
```
huitseeker, line 15: This function looks great, though I'm not sure about its name: ell isn't a standard nomenclature.
Here are things that would help me grok what this function is for out of context, and that you could use to edit the comment, the name of this function, or more likely both :
- make it clear we're testing the prove and verify flow of *Multilinear* Polynomial Commitment Schemes (PCS), 
- which in Nova are represented by `EvaluationEngineTrait` implementations for proving, 
- replacing `ell` by the more evocative `num_var`, which becomes clear once the first bullet point has been stated clearly,
- avoid the mention of an `Engine`, which is implied and selected by a choice of `EvaluationEngineTrait` implementation,

src/provider/non_hiding_zeromorph.rs
```diff
@@ -443,19 +443,19 @@ fn eval_and_quotient_scalars<F: Field>(y: F, x: F, z: F, point: &[F]) -> (F, (Ve
...
-    .zip(&vs)
-    .zip_eq(&vs[1..])
-    .zip_eq(point.iter().rev()) // assume variables come in BE form
-    .map(
-      |(((((power_of_y, offset_of_x), square_of_x), v_i), v_j), u_i)| {
-        (-(power_of_y * offset_of_x), -(z * (square_of_x * v_j - *u_i * v_i)))
-      },
-    )
-    .unzip();
+      .zip_eq(offsets_of_x)
```
huitseeker, line 446: Weird that `cargo fmt` didn't complain with the original indentation!?

src/spartan/polys/multilinear.rs
```diff
@@ -63,8 +63,12 @@ impl<Scalar: PrimeField> MultilinearPolynomial<Scalar> {
     self.Z.len()
   }
 
+  /// Returns true if no evaluations.
```
huitseeker, line 66: Note for posterity: removed from upstream as unused code in https://github.com/microsoft/Nova/pull/257, I agree with its re-introduction here because of [len_without_is_empty](https://rust-lang.github.io/rust-clippy/master/index.html#/len_without_is_empty)

huitseeker (commented):
This mostly looks great, I left some inline comments for simplification.
A rebase would be great, as it would allow us to see improvements from #216.

huitseeker (commented):
One issue that only pops off the page now: at large values of ell, the generation of a proof is expensive, as you've no doubt noticed. Nonetheless, the output is completely deterministic.

One bad thing that's obscured by the `benchmark_engines` macro is that you're generating assets for a specific size (and EE choice), benching the proving, then re-generating those assets for the verification, and finally benching the verification. The assets could and should be shared between proving and verification I think.

benches/pcs.rs
```diff
@@ -0,0 +1,181 @@
...
+          name = pcs;
+          config = Criterion::default().warm_up_time(Duration::from_millis(3000));
+          targets = bench_pcs
+    }
+  }
+}
+
+criterion_main!(pcs);
+
+const TEST_ELL: [usize; 11] = [10, 11, 12, 23, 14, 15, 16, 17, 18, 19, 20];
```
huitseeker, line 38: I'm not sure we need all points in the [10-20] interval. [10, 12, 14 ,16, 18, 20] seem amply sufficient to me.

src/provider/non_hiding_zeromorph.rs
tchataigner: Yes pretty strange, I kept the update as it seemed ok

benches/pcs.rs
```diff
@@ -0,0 +1,181 @@
...
+          (Bn256EngineKZG, MLEvaluationEngine<Bn256, Bn256EngineKZG>, "MLKZG"),
+          (Bn256EngineZM, ZMPCS<Bn256, Bn256EngineZM>, "ZM")
+      );
+
+      verifying_group.finish();
+    }
+  }
+}
+
+fn deterministic_assets<E: Engine, EE: EvaluationEngineTrait<E>, R: CryptoRng + RngCore>(
```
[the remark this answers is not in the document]
tchataigner: You're right that the naming here is off. I specified it at `deterministic...` because if was a pre-determined polynomial size used to generate the assets. I've the implementation to be over `BenchAssets` and the name of the method to `from_num_vars`. should be more explicit this way.

src/spartan/polys/multilinear.rs
```diff
@@ -73,6 +77,27 @@ impl<Scalar: PrimeField> MultilinearPolynomial<Scalar> {
     )
   }
 
+  /// Returns a random polynomial, a point and calculate its evaluation.
+  pub fn random_with_eval<R: RngCore + CryptoRng>(
```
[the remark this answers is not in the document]
tchataigner: This was actually an update on my code based on a discussion with @adr1anh that pointed out to me that if we had a `random` we could have a `random_with_eval`.

What you're saying about it being a test function is true, and it is also the same for `random`. I could remove both from the public API and instead inline their code where we have some usage.

Would that be alright with you?

src/provider/util/mod.rs
```diff
@@ -1,3 +1,97 @@
...
+    assert_eq!(pcp, pcv);
+
+    if evaluate_bad_proof {
+      // Generate another point to verify proof. Also produce eval.
+      let altered_verifier_point = point
+        .iter()
+        .map(|s| s.add(<E as Engine>::Scalar::ONE))
+        .collect::<Vec<_>>();
+      let altered_verifier_eval =
+        MultilinearPolynomial::evaluate_with(poly.evaluations(), &altered_verifier_point);
```
adr1anh, line 82: Instead of running the verification twice, you can do something like 
```rust 
let point = if evaluate_bad_proof {
    altered_verifier_point
} else {
  point
}
// Verify proof, should fail.
let mut verifier_transcript = E::TE::new(b"TestEval");
let res = EE::verify(
        &verifier_key,
        &mut verifier_transcript,
        commitment,
        &altered_verifier_point,
        &altered_verifier_eval,
        &proof,
);
assert_eq!(res.is_err(), evaluate_bad_proof); 
```

tchataigner: That's true, however this would also change how the method behave. Currently if `evaluate_bad_proof` is set to `true` it allows the method to test both a valid and a non valid proof. While your proposal makes it that I only test one of the two cases. The downside is that I would then have to call this helper method twice for unit tests, doubling the number of `prove` method call instead.

Even after this comment, if you still think I should go with your proposal I'll be happy to do it :)

benches/pcs.rs
```diff
@@ -0,0 +1,181 @@
...
+
+fn bench_pcs(c: &mut Criterion) {
+  for &ell in TEST_ELL.iter() {
+    let mut rng = rand::rngs::StdRng::seed_from_u64(ell as u64);
+
+    // Proving group
+    {
+      let mut proving_group = c.benchmark_group("PCS-Proving");
+      proving_group
+        .sampling_mode(SamplingMode::Flat)
```
storojs72, line 71: As far as I understand, `Flat` is only suitable to use for very long-running benchmarks. Probably `Auto` can be more appropriate for the short-running ones, where `ell` is say 10 / 11 / 12. Does it make sense to split the benchmarks into short / long running categories?

benches/pcs.rs
```diff
@@ -0,0 +1,181 @@
...
+    &prover_key,
+    &mut E::TE::new(b"TestEval"),
+    &commitment,
+    poly.evaluations(),
+    &point,
+    &eval,
+  )
+  .unwrap();
+
+  BenchAssests {
```
storojs72, line 134: BenchAssets?

storojs72 (commented):
LGTM!

Obviously, it is hard to make some statements about performance without running benchmarks on variety of ELLs but I would may be consider specifying the `ELL` as a single digit via a command-line which can be useful while running single instance of the benchmark on developer's machine in order to get very rough information whether performance has been increased / decreased / left unchanged after some PR merged.

benches/pcs.rs
huitseeker: Funny enough: in general the `Flat` sampling mode is fine for most cases ... except I looked at the sample distribution for this PR, and the standard deviation for ZM verification at ell = 10 doesn't look great. You're probably right @storojs72, good catch!

Though I would note one very simple way of addressing this is noting the bench runtimes, which are reasonable (<20s per ell setting, see the bench log on my M2 at https://gist.github.com/huitseeker/015f07534f410e94d615684ecd86876f) re-run them *without* setting the sampling mode (defaulting to `Auto`), and see if the runtimes are still reasonable.

Screenshot 2024-01-04 at 10 03 38 AM
Screenshot 2024-01-04 at 10 04 06 AM
Screenshot 2024-01-04 at 10 04 22 AM
Screenshot 2024-01-04 at 10 04 30 AM

huitseeker:
> Obviously, it is hard to make some statements about performance without running benchmarks on variety of ELLs but I would may be consider specifying the ELL as a single digit via a command-line which can be useful while running single instance of the benchmark on developer's machine in order to get very rough information whether performance has been increased / decreased / left unchanged after some PR merged.

A way to do this is to have the ELL array (currently a constant) be overridden by an environment variable. We can probably open an issue and tackle this after merging this PR.

benches/pcs.rs
tchataigner: Assets used during our benchmark that are shared between the proving and verification steps. Useful to avoid having lengthy methods signatures.

benches/pcs.rs
tchataigner: As we have removed all odds sizes we are only left with 10 and 12 from your example. I'm not sure that there is a need to split the code for 2 cases. 

I would either make it `Auto` for all (might not be great for higher-end number of variables) or `Flat` (losing a bit of iterations on smaller number of variables but we should still get enough to have convincing results).

huitseeker: I'm explicitly suggesting Auto for all, and then revisiting the question if the overall bench runtime is larger than say 15 min.

tchataigner: I don't know why I couldn't see your comment when I put mine up. I'll update to `Auto`

src/spartan/polys/multilinear.rs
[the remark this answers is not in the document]
tchataigner: Moved `random_with_eval` only to eval :)

huitseeker (approved):
I've re-run the benches on the same machine and the end-to-end runtime stands at 00:16:30, which I deem acceptable.
This looks great @tchataigner, thank you so much!

src/provider/non_hiding_zeromorph.rs
huitseeker: I was wrong, you were right, I misread your test utils 🤦

PR #143, opened 2023-11-27
Preallocate witness before synthesis

winston-h-zhang:
Addresses #137 and replaces #138.

This PR uses the new API in `WitnessCS` to preallocate the witness buffers before calling `synthesize`. This ensures that we do not unnecessarily grow and reallocate these large witness buffers, which adds overhead.

PR #139, opened 2023-11-27
Reduce allocations/cloning in spmvm

winston-h-zhang:
Addresses #137.

This PR refactors the `R1CSShape` sparse matrix / dense vector multiplication API to avoid the unnecessary cloning and allocating of memory. In the codebase, we often see the pattern:
```rust
let z = [W.W.clone(), vec![U.u], U.X.clone()].concat();
let (Az, Bz, Cz) = self.multiply_vec(&z)?;
```
This is memory-inefficient. The new API directly uses references to the witness vectors, like this:
```rust
let (Az, Bz, Cz) = self.multiply_witness(&W.W, &U.u, &U.X)?;
```

Benchmark results from downstream `lurk-rs` confirm that this improves performance: https://github.com/lurk-lab/lurk-rs/pull/923

## Pasted Benchmark
  <details>
    <summary>Click to view benchmark</summary>

| Test | Base         | PR               | % |
|------|--------------|------------------|---|
| LEM Fibonacci Prove - rc = 100/fib/num-100 | 4.8±0.03s | **4.4±0.03s** | **-8.33%** |
| LEM Fibonacci Prove - rc = 100/fib/num-200 | 9.7±0.33s | **9.2±0.08s** | **-5.15%** |
| LEM Fibonacci Prove - rc = 600/fib/num-100 | 3.8±0.01s | N/A | N/A |
| LEM Fibonacci Prove - rc = 600/fib/num-200 | 8.4±0.10s | N/A | N/A |

  </details>

PR #227, opened 2024-01-03
Cleanup memory pipeline for SuperNova

winston-h-zhang:
Resolves the next step of #137.

This PR brings the changes done to the Nova memory pipeline in #139, #143, and #144 to SuperNova as well. Since the core logic was already established in the previous Nova PRs, this PR mostly just refactors and cleans up the SuperNova API.

Notes:
- We don't remove the use of `R1CSShape::multiply_vec`, since it's still in `(pp)snark.rs`. The concatenation of `z = W.W ++ U.u ++ U.X` happens anyways, so we can keep the instances of `S.multiply_vec(z)` for readability.

### DIFF
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -703,11 +703,10 @@ where
     )
     .map_err(SuperNovaError::NovaError)?;
 
-    // // clone and updated running instance on respective circuit_index
-    // let r_U_secondary_next = self.r_U_secondary;
-    // let r_W_secondary_next = self.r_W_secondary;
-
-    let mut cs_primary = SatisfyingAssignment::<E1>::new();
+    let mut cs_primary = SatisfyingAssignment::<E1>::with_capacity(
+      pp[circuit_index].r1cs_shape.num_io + 1,
+      pp[circuit_index].r1cs_shape.num_vars,
+    );
     let T =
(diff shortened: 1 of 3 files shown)


src/supernova/mod.rs
```diff
@@ -593,6 +614,31 @@ where
...
+      }
+    });
+
+    let buffer_primary = ResourceBuffer {
+      l_w: None,
+      l_u: None,
+      ABC_Z_1: R1CSResult::default(max_r1cs_primary),
+      ABC_Z_2: R1CSResult::default(max_r1cs_primary),
+      T: r1cs::default_T(max_r1cs_primary),
+    };
```
adr1anh, line 632: `R1CSResult::default()` only requires the `num_cons` value from the `R1CSShape`. If we change the function to take this value instead, then we can more simply compute

```rust
let max_num_cons = *pp.circuit_shapes.iter().map(|shape| shape.num_cons).max().unwrap();
let buffer_primary = ResourceBuffer {
      l_w: None,
      l_u: None,
      ABC_Z_1: R1CSResult::default(max_num_cons),
      ABC_Z_2: R1CSResult::default(max_num_cons),
      T: r1cs::default_T(max_num_cons),
    }; 
```

adr1anh (approved):
Small nit about initializing the `ResourceBuffer` but other than that it looks good to me!

huitseeker (approved):
This looks great, thank you so much!

PR #238, opened 2024-01-05
ci: Add comparative GPU benchmarks on `merge_group`

samuelburnham:
# GPU benchmarks for PR regression testing
- Adds a benchmark regression test when merging a PR, based on https://github.com/lurk-lab/lurk-rs/blob/main/.github/workflows/merge-tests.yml
- Refactors the `recursive-snark` benchmark group and ID to enable formatting with `criterion-table`. This will be extended to `compressed-snark` and their supernova counterparts when ready.
- I attempted to test the `criterion-table` output in https://github.com/lurk-lab/arecibo/commit/a621594fa19c6ba89bd93c0bd3772f3f2765c6b2#commitcomment-136557829, and the basic logic works. Unfortunately, we're limited by the fact that the base benchmarks on `dev` don't have the same formatting or parameters as this PR, so it's impossible to demo exactly what it will look like beyond the proposal below.
  - It does correctly pick up the artificial regression I introduced, which opened #245 (partially thanks to the Dockerfile fix in https://github.com/lurk-lab/gh-actions-runner/pull/25)
- I also can't run these actions in a fork, so rather than a suboptimal reconstruction in https://github.com/lurk-lab/ci-lab I'd like to merge this PR and then iron out any last issues in the interests of speed.

Proposed `criterion-table` output:

# Benchmarks

## Table of Contents

- [Benchmark Results](#benchmark-results)
    - [RecursiveSNARK](#recursivesnark)

### RecursiveSNARK

|                                   | `ref=ae521be709871f94a622c85b0e5d6e8f776be144`          | `ref=ace6724dc77667506dbcd1122b9851bd6edb25d2`           |
|:----------------------------------|:--------------------------------------------------------|:-------------------------------------------------------- |
| **`Prove-StepCircuitSize-0`**     | `29.28 ms` (✅ **1.00x**)                                | `28.71 ms` (✅ **1.02x faster**)                          |
| **`Verify-StepCircuitSize-0`**    | `19.48 ms` (✅ **1.00x**)                                | `19.10 ms` (✅ **1.02x faster**)                          |
| **`Prove-StepCircuitSize-6559`**  | `35.57 ms` (✅ **1.00x**)                                | `35.27 ms` (✅ **1.01x faster**)                          |
| **`Verify-StepCircuitSize-6559`** | `23.84 ms` (✅ **1.00x**)                                | `23.89 ms` (✅ **1.00x slower**)                          |

---
Made with [criterion-table](https://github.com/nu11ptr/criterion-table)

## Design notes
For the other benchmarks which have more than one group, they would each appear in their own table (e.g. `CompressedSNARK` and `CompressedSNARK-Commitments`). This layout will be shown as well when implemented.

Some other display options:
- Split each `StepCircuitSize` into its own group, which is the existing behavior. This would proliferate the number of benchmark tables to a large degree if we are displaying four different sets of benchmarks in one `.md` file
- Split the `Prove` and `Verify` benchmarks into their own table or group them together. This would require some finagling since the benchmarks are currently ordered sequentially in the order they are run.

## Future work
- Refactor into a reusable workflow (https://github.com/lurk-lab/ci-workflows/issues/15)
- Cache the PR benchmark for reuse (#246)

Helps #243

### DIFF
--- a/.github/PERF_REGRESSION.md
+++ b/.github/PERF_REGRESSION.md
@@ -2,6 +2,6 @@
 title: ":rotating_light: Performance regression in #{{ env.PR_NUMBER }}"
 labels: P-Performance, automated issue
 ---
-Regression >= 5% found during merge of: #{{ env.PR_NUMBER }}
+Regression >= {{ env.NOISE_THRESHOLD }} found during merge of: #{{ env.PR_NUMBER }}
 Commit: {{ env.GIT_SHA }}
 Triggered by: {{ env.WORKFLOW_URL }}
--- a/.github/PERF_REGRESSION.md
+++ b/.github/PERF_REGRESSION.md
@@ -1,0 +1,7 @@
+---
+title: ":rotating_light: Performance regression in #{{ env.PR_NUMBER }}"
+labels: P-Performance, automated issue
+---
+Regression >= 5% found during merge of: #{{ env.PR_NUMBER }}
+Commit: {{ env.GIT_SHA }}
+Triggered by: {{ env.WORKFLOW_URL }}
--- a/.github/tables.toml
+++ b/.github/tables.toml
@@ -1,0 +1,6 @@
+[table_comments]
+
+[top_comments]
+Overview = """
+This benchmark report shows the Arecibo GPU benchmarks.
+"""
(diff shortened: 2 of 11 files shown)


huitseeker (approved):
Agreed, this looks like a lot of progress, let's merge and iron out the kinks.

Issue #228, opened 2024-01-03
Avoid computing polynomial which is constant


storojs72:
Addresses [following](https://github.com/lurk-lab/arecibo/pull/172#discussion_r1427834586) comment from @adr1anh 

Essentially we can save some computations by exploiting the fact that final polynomial in `polys` equals to the `eval`. So we don't need to compute it during [following](https://github.com/lurk-lab/arecibo/blob/dev/src/provider/mlkzg.rs#L238) loop and avoid committing to it - since it is cheaper to derive the commitment on verification side

PR #242, opened 2024-01-08
Optimising proving by avoiding recomputing final `Pi` polynomial in MLKZG

storojs72:
Fixes https://github.com/lurk-lab/arecibo/issues/228

The obvious solution could be just reducing the number of iteration in [following]((https://github.com/lurk-lab/arecibo/blob/dev/src/provider/mlkzg.rs#L232)) loop to `ell-1` with subsequent pushing `eval` to `polys`. But, as @adr1anh noted in some of discussions we don't need to compute the commitment to this constant polynomial at prover side, since verifier has all necessary information to do that on its side, so in this PR, the `EvaluationArgument` created by prover contains `comms` without commitment to `eval`. It is recomputed at verifier's side and added to `com` vector before executing `kzg_verify_batch`. 

**Important detail**

The following check:
```
if i == ell - 1 && *eval != Pi[0] {
    return Err(NovaError::UnSat);
}
```

from initial implementation is no longer needed, because we don't compute final `Pi` but add `eval` to `polys` instead, so it is removed. The wrong `eval` is now detected during proof verification and unit-test added to enforce this. 

**Question to reviewers**

We don't actually use the `C` parameter while proving. In original implementation it [was used](https://github.com/lurk-lab/arecibo/blob/dev/src/provider/mlkzg.rs#L197) only inside `kzg_open_batch` just to validate the input size and nowhere else at sensible computations. Is that expected?

### DIFF
--- a/src/provider/mlkzg.rs
+++ b/src/provider/mlkzg.rs
@@ -164,10 +164,9 @@ where
         .to_affine()
     };
 
-    let kzg_open_batch = |C: &[E::G1Affine],
-                          f: &[Vec<E::Fr>],
+    let kzg_open_batch = |f: &[Vec<E::Fr>],
                           u: &[E::Fr],
-                          transcript: &mut <NE as NovaEngine>::TE|
+                          _transcript: &mut <NE as NovaEngine>::TE|
      -> (Vec<E::G1Affine>, Vec<Vec<E::Fr>>) {
       let scalar_vector_muladd = |a: &mut Vec<E::Fr>, v: &Vec<E::Fr>, s: E::Fr| {
         assert!(a.len() >= v.len());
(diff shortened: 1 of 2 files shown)


src/provider/mlkzg.rs
```diff
@@ -294,6 +289,7 @@ where
       let k = C.len();
       let t = u.len();
 
+      //let q = Self::get_batch_challenge_ignoring_eval(v, eval, transcript);
```
adr1anh, line 292: Can we remove this?

src/provider/mlkzg.rs
```diff
@@ -263,9 +259,8 @@ where
     let u = vec![r, -r, r * r];
 
     // Phase 3 -- create response
-    let mut com_all = comms.clone();
-    com_all.insert(0, C.comm.to_affine());
-    let (w, evals) = kzg_open_batch(&com_all, &polys, &u, transcript);
+    polys.push(vec![*eval]); // final 'Pi' polynomial is constant and equals to eval
```
adr1anh, line 262: We don't need to add the constant polynomial here.

src/provider/mlkzg.rs
```diff
@@ -380,12 +376,15 @@ where
...
-
     if r == E::Fr::ZERO || C.comm == E::G1::identity() {
       return Err(NovaError::ProofVerifyError);
     }
     com.insert(0, C.comm.to_affine()); // set com_0 = C, shifts other commitments to the right
 
+    // compute commitment to constant eval and add it commitments vector
+    let com_to_eval = E::G1Affine::from(vk.g * (*y));
+    com.push(com_to_eval);
+
```
adr1anh, line 387: We don't need to create a commitment to the constant polynomial here

adr1anh (commented):
This looks correct to me! However, we can be slightly more optimal by avoiding the opening of the constant polynomial entirely. 

The prover only needs to open the `ell` polynomials, where the first one is the initial input one, and the last one is of size 2. In the first phase of the verification, the verifier uses the fact that `ypos[ell] = yneg[ell] = Y[ell] = *y` (the evaluation of the constant polynomial). 

- Change the size to `ell` rather than `ell+1` in [L395](https://github.com/lurk-lab/arecibo/pull/242/files#diff-2a9a0f2c0544683992c509e0ce2607d1bf88f35943ac0f735ca25ee749eb14e6R395-R397)
- Append `*y` to `Y` in [L400](https://github.com/lurk-lab/arecibo/pull/242/files#diff-2a9a0f2c0544683992c509e0ce2607d1bf88f35943ac0f735ca25ee749eb14e6R400).

adr1anh (approved):

PR #253, opened 2024-01-10
Refactor Supernova benchmarks and config for CI

samuelburnham:
This PR refactors the Supernova benchmarks to reduce duplicated code and enables configuration of the constraint numbers used in the benchmarks via the `ARECIBO_NUM_CONS` env var.

- Factors out common functionality into `benches/common/supernova`. The `bench_snark_internal_with_arity` function is a bit hacky, as the `S1, S2` trait bounds are required but unused by the recursive SNARK. There is surely a better trait solution but this seemed simplest for now.
- Creates a new `supernova-ci` benchmark to run the `RecursiveSNARKSuperNova-2circuit` and `CompressedSNARKSuperNova-Commitments-2circuit` with 16384/1038732 constraints to get a good idea of perf changes on each PR, and updates CI accordingly.
  -  I tried to configure the benchmark targets directly by `quote!`-ing a vector of strings into a list of tokens and passing that to the `criterion_group!` targets, but it expected a `path` specifier instead of an `expr` so this was unsuccessful. Other than this I couldn't think of a way to configure which targets to run besides feature configuration or the implemented solution of creating a new benchmark.
- Sets the noise threshold to 0.30 to reduce noise (e.g. #248, #251)


benches/common/supernova/bench.rs
```diff
@@ -0,0 +1,179 @@
...
+pub fn run_bench<S1: BatchedRelaxedR1CSSNARKTrait<E1>, S2: RelaxedR1CSSNARKTrait<E2>>(
+  c: &mut Criterion,
+  group_name: &str,
+  arity: usize,
+  snark_type: SnarkType,
+) {
+  // we vary the number of constraints in the step circuit
+  for &num_cons_in_augmented_circuit in num_cons().iter() {
+    // number of constraints in the step circuit
+    let num_cons = num_cons_in_augmented_circuit - NUM_CONS_VERIFIER_CIRCUIT_PRIMARY;
```
huitseeker, line 169: Yeah, so this is the problem: in bench mode, that substraction has wrapping semantics, and if a missed update of the cons numbers makes it wrap around, the benches will silently start eating up the bench machine's memory. We should do `checked_sub(..).expect("constraint numbers out of date!!")` here

benches/common/supernova/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+pub type S1 = arecibo::spartan::batched::BatchedRelaxedR1CSSNARK<E1, EE1>;
+pub type S2 = arecibo::spartan::snark::RelaxedR1CSSNARK<E2, EE2>;
+// SNARKs with computation commitments
+pub type SS1 = arecibo::spartan::batched_ppsnark::BatchedRelaxedR1CSSNARK<E1, EE1>;
+pub type SS2 = arecibo::spartan::ppsnark::RelaxedR1CSSNARK<E2, EE2>;
+
+// This should match the value in test_supernova_recursive_circuit_pasta
+// TODO: This should also be a table matching the num_augmented_circuits in the below
+pub const NUM_CONS_VERIFIER_CIRCUIT_PRIMARY: usize = 9844;
+pub const NUM_SAMPLES: usize = 10;
```
huitseeker, line 31: Not-so-fun part: this number is meant to be meaningfully different for the nova and supernova use cases. We may want to document that.
This number is also a footgun waiting to happen, see my note on `let num_cons = ...`

huitseeker (commented):
Very encouraging, and I'm ok with the type hackery for now.

PR #257, opened 2024-01-12
fix: Integrate CUDA feature with grumpkin_msm dependency

huitseeker:
- Modified grumpkin-msm dependency to be optional for x86_64/aarch64 targets and added it to CUDA dependencies in `Cargo.toml`.
- Inserted `cuda` feature guard for `vartime_multiscalar_mul` function in `traits.rs` file to only allow execution when `cuda` feature and specific architecture are enabled.
- Adjusted targeting condition for grumpkin_msm algorithm use in `pasta.rs`. The algorithm is now dependent on the 'cuda' feature and x86_64 or aarch64 architecture.
- Updated conditional compilation in `bn256_grumpkin.rs` to integrate CUDA feature for both bn256 and grumpkin.

Fixes https://github.com/lurk-lab/lurk-rs/issues/1030

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ derive_more = "0.99.17"
 [target.'cfg(any(target_arch = "x86_64", target_arch = "aarch64"))'.dependencies]
 # grumpkin-msm has been patched to support MSMs for the pasta curve cycle
 # see: https://github.com/lurk-lab/grumpkin-msm/pull/3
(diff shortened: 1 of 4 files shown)


huitseeker:
After benchmarking, the CPU performance of `cpu_best_msm` is insufficient (relative to what grumpkin-msm will provide on the CPU) to make this switch.

PR #265, opened 2024-01-16
Shorten benchmark names and fix CI

samuelburnham:
- Shortens Criterion benchmark report names to account for the 100 character limit in https://github.com/bheisler/cargo-criterion/blob/main/src/report.rs#L123. Starts with Supernova benches as the others aren't used in CI
- Fixes a missing input to `criterion-table` in the GPU regression check


huitseeker (approved):
🙏

PR #268, opened 2024-01-17
refactor: Remove clone trait from key and SNARK structs

huitseeker:
## Summary

We've been a bit loose in allowing macro-derived Clone implementations in spartan SNARK structures. This removes the unused ones, and checks that we won't re-introduce these Clone implementations in the future for structs where this will not make sense (`ProverKey`, `VerifierKey`, `SNARK`).

- After https://github.com/lurk-lab/lurk-rs/pull/1064

## Details
- Removed `Clone` trait from various structs across multiple files, including `ProverKey`, `VerifierKey`, and multiple versions of `RelaxedR1CSSNARK` and `BatchedRelaxedR1CSSNARK`, among others.
- Instituted tests to confirm that the `Clone` trait is **not** implemented for previously mentioned structs where Clone will not make sense.
- Added a new `static_assertions` dependency to the project configuration file, Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ itertools = "0.12.0"
 rand = "0.8.5"
 ref-cast = "1.0.20"
 derive_more = "0.99.17"
+static_assertions = "1.1.0"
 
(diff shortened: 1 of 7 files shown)


huitseeker:
Shelving until I get time to come back to it.

PR #269, opened 2024-01-17
feat: Implement batch operations in non_hiding_kzg module

huitseeker:
- Added a `batch_commit` method to generate multiple commitments from a list of polynomials.
- Introduced a `batch_open` functionality for evaluating multiple polynomials at different points.
- Implemented `batch_verify` function for validation of polynomial evaluations in a multi-commitment setup.
- Verified the correctness of the batch operations with a new unit test `batch_check_test`.

These basic variants of batch KZG were originally part of #145

### DIFF
--- a/src/provider/non_hiding_kzg.rs
+++ b/src/provider/non_hiding_kzg.rs
@@ -4,6 +4,7 @@ use ff::{Field, PrimeField, PrimeFieldBits};
 use group::{prime::PrimeCurveAffine, Curve, Group as _};
 use pairing::{Engine, MillerLoopResult, MultiMillerLoop};
 use rand_core::{CryptoRng, RngCore};
+use rayon::iter::{IntoParallelIterator, ParallelIterator};
 use serde::{Deserialize, Serialize};
 use std::{borrow::Borrow, marker::PhantomData, ops::Mul};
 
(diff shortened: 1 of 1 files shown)


src/provider/non_hiding_kzg.rs
```diff
@@ -334,6 +376,61 @@ where
...
+        randomizer = E::Fr::from_u128(rand::Rng::gen::<u128>(rng));
+      }
+    );
+    total_c -= &verifier_params.g.mul(g_multiplier);
+
+    let mut affine_points = vec![E::G1Affine::identity(); 2];
+    E::G1::batch_normalize(&[-total_w, total_c], &mut affine_points);
+    let (total_w, total_c) = (affine_points[0], affine_points[1]);
+
+    let result = E::multi_miller_loop(&[
```
storojs72, line 424: As Adrian mentioned (and I then could also convince myself by observing the implementation), this form of pairings check is more optimal, comparing to:

```
let left = E::pairing(..., ...);
let right = E::pairing(..., ...);
assert_eq!(left, right);
```

as it saves one `multi_miller_loop` and one `final_exponentiation` call. However that requires some manipulating with input parameters, which differs from what is necessary to pass to `pairing` couple. For example, given vanilla KZG10 specification, described [here](https://risencrypto.github.io/Kate/), the final check is specified like:

```
e(C_q, a⋅G2−b⋅G2) == e(C_f−G1⋅c, G2) 
```

which has following Rust equivalent:
```
let g1 = srs.powers_of_g[0];
let g2 = srs.powers_of_h[0];

let left = pairing(
  &Cq.to_affine(),
  &(srs.powers_of_h[1] - (g2 * b)).to_affine(),
);
let right = pairing(&(Cf.to_affine() - (g1 * c).to_affine()).to_affine(), &g2);
assert_eq!(left, right);
```

Alternatively (more optimally), one can write this check as:
```
// Alternatively (more optimal as it doesn't require executing final_exponentiation twice)
let term1 = (
  &(srs.powers_of_g[0] * c - Cq * b - Cf).to_affine(),
  &G2Prepared::from_affine(srs.powers_of_h[0]),
);
let term2 = (
  &Cq.to_affine(),
  &G2Prepared::from_affine(srs.powers_of_h[1]),
);
let result = multi_miller_loop(&[term1, term2]).final_exponentiation();
assert_eq!(result.is_identity().unwrap_u8(), 0x01);
```
which I derived empirically. 

I wonder if there is a systematic way that allows turning the `e() == e()` formula into more optimal one:
```
term1 = ...
term2 = ...
multi_miller_loop(term1, term2).final_exponentiation().is_identity()
```

Thanks in advance, for answering, @huitseeker !

storojs72 (approved):

PR #273, opened 2024-01-19
CycleFold

mpenciak:
This PR is a WIP for CycleFolding with Nova IVC.

TODO:

- [ ] ~~Incorporate the changes from https://github.com/microsoft/Nova/pull/291 in this `RecursiveSNARK` implementation~~ (The same assumptions in this PR do not hold for the CycleFold folding, so it would be unsound to include these changes)
- [ ] ~~Incorporate the changes from https://github.com/microsoft/Nova/pull/295 in this `RecursiveSNARK` implementation~~


src/cyclefold/circuit.rs
```diff
@@ -0,0 +1,222 @@
...
+      result,
+      scalar,
+    }
+  }
+}
+
+/// TODO: docs
+pub struct CyclefoldCircuit<E: Engine> {
+  inputs: Option<CyclefoldCircuitInputs<E>>,
+}
```
porcuquine, line 39: What's the benefit of this indirection? Why not just implement `synthesize`, etc. for the 'inputs'?

porcuquine: I get that you're wrapping in an `Option`, but this is already accomplished by the internal `Option`s in the contained struct. Moreover, the field is a `Vec` of `Option`s, and even when synthesizing shape, this should still be manifested in order to make the size of the inputs explicit, etc.

I maybe missing something, but I would probably reduce this to a single struct, perhaps with an initializer making it easy to create a (correctly-sized) empty version; and choose a name reflecting the result.

src/cyclefold/circuit.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    let scalar = self
+      .inputs
+      .as_ref()
+      .ok_or(SynthesisError::AssignmentMissing)?
+      .scalar
+      .iter()
+      .enumerate()
+      .map(|(idx, b)| AllocatedBit::alloc(cs.namespace(|| format!("scalar bit {}", idx)), *b))
+      .collect::<Result<_, _>>()?;
```
porcuquine, line 91: This looks like it will only allocate the bits when the inputs are present. Won't this result in non-determinism in the shape of the circuit (and hence lead to unverifiable proofs) or other runtime errors?

src/cyclefold/circuit.rs
```diff
@@ -0,0 +1,222 @@
...
+      |lc| lc + CS::one(),
+      |lc| lc + C_final.y.get_variable(),
+    );
+
+    cs.enforce(
+      || "C_final_inf = res_inf",
+      |lc| lc + res_inf.get_variable(),
+      |lc| lc + CS::one(),
+      |lc| lc + C_final.is_infinity.get_variable(),
+    );
```
porcuquine, line 130: It would simplify confidence in correctness if you implemented `AllocatedPoint::enforce_equal` or similar, then used it throughout. Manually writing out all these constraints introduces many opportunities for small errors or inconsistencies, and at best increases the review/audit burden.

porcuquine: Even barring that, you could do similar for the components rather than manually constructing all the constraints here.

src/cyclefold/circuit.rs
```diff
@@ -0,0 +1,222 @@
...
+    let P_x = P.x.get_value().unwrap();
+    let P_y = P.y.get_value().unwrap();
+    let P_infty = P.is_infinity.get_value().unwrap();
+
+    let (x, y, infty) = C_final.to_coordinates();
+
+    assert_eq!(x, P_x);
+    assert_eq!(y, P_y);
+    assert_eq!(infty, P_infty == <E1::Scalar as Field>::ONE);
+  }
```
porcuquine, line 214: You should probably check for circuit satisfaction in the test also.

src/cyclefold/gadgets.rs
```diff
@@ -0,0 +1,614 @@
...
+          Ok(f_to_nat(
+            &coords.ok_or(SynthesisError::AssignmentMissing)?.1,
+          ))
+        },
+        limb_width,
+        n_limbs,
+      )?;
+
+      let is_infinity = AllocatedNum::alloc(cs.namespace(|| "is_infinity"), || {
+        Ok(if coords.map_or(true, |c| c.2) {
```
porcuquine, line 130: Why do you silently map a missing infinity coordinate to `true` here rather than return an `AssignmentMissing` as elsewhere (and expected)?

src/cyclefold/gadgets.rs
```diff
@@ -0,0 +1,614 @@
...
+        } else {
+          E1::Base::ZERO
+        })
+      })?;
+      cs.enforce(
+        || "is_infinity is bit",
+        |lc| lc + is_infinity.get_variable(),
+        |lc| lc + CS::one() - is_infinity.get_variable(),
+        |lc| lc,
+      );
```
porcuquine, line 141: Why do you not just allocate an `AllocatedBit` instead of manually simulating one?

src/cyclefold/gadgets.rs
```diff
@@ -0,0 +1,614 @@
...
+        limb_width,
+        n_limbs,
+      )?;
+
+      let B_bn = alloc_bignat_constant(
+        cs.namespace(|| "alloc B"),
+        &f_to_nat(&E1::GE::group_params().1),
+        limb_width,
+        n_limbs,
+      )?;
```
porcuquine, line 215: It would be better to allocate these constants once (if they really need to be allocated at all) rather than every time `check_on_curve` is called.

src/cyclefold/gadgets.rs
```diff
@@ -0,0 +1,614 @@
...
+
+      cs.enforce(
+        || "always_equal = 1 - is_infinity",
+        |lc| lc,
+        |lc| lc,
+        |lc| lc + always_equal.get_variable() - CS::one() + self.is_infinity.get_variable(),
+      );
+
+      y_sq.equal_when_carried_regroup(cs.namespace(|| "y_sq = rhs"), &rhs, &always_equal)?;
```
porcuquine, line 241: Please write out the equation being checked here in a clear comment.

src/gadgets/nonnative/bignat.rs
```diff
@@ -354,10 +369,11 @@ impl<Scalar: PrimeField> BigNat<Scalar> {
     &self,
     mut cs: CS,
     other: &Self,
+    always_equal: &AllocatedBit,
```
porcuquine, line 372: Please document the semantics of this parameter.

src/cyclefold/gadgets.rs
```diff
@@ -0,0 +1,586 @@
...
+      },
+    },
+    traits::{CurveCycleEquipped, Dual, Group, ROConstantsCircuit},
+    RelaxedR1CSInstance,
+  };
+
+  use ff::Field;
+
+  #[derive(Clone)]
+  pub struct AllocatedPoint<E>
```
adr1anh, line 80: I think we should avoid the `is_infinity` bit here. In order to properly constrain it, we need to evaluate the curve equation non-natively which will be very expensive. Instead, (and if the properties of the curve allow), we could define `(0,0)` as the identity and let the CycleFold circuit check that this holds.

src/gadgets/nonnative/bignat.rs
```diff
@@ -438,6 +454,7 @@ impl<Scalar: PrimeField> BigNat<Scalar> {
...
```
adr1anh, line 457: Making this a `Boolean` and performing the conversion from `AllocatedBit` allows the function to skip the additional constraint when `always_equal` is a `Boolean::Constant`.

src/cyclefold/nova_circuit.rs
```diff
@@ -0,0 +1,565 @@
...
+    )?;
+
+    Ok((U_c, U_p, checks_pass))
+  }
+
+  pub fn synthesize<CS: ConstraintSystem<E::Base>>(
+    self,
+    cs: &mut CS,
+  ) -> Result<Vec<AllocatedNum<E::Base>>, SynthesisError> {
+    // TODO: It's written down here https://hackmd.io/@mpenciak/HybHrnNFT
```
adr1anh, line 415: Broken link

src/cyclefold/circuit.rs
```diff
@@ -0,0 +1,222 @@
...
+      .iter()
+      .enumerate()
+      .map(|(idx, b)| AllocatedBit::alloc(cs.namespace(|| format!("scalar bit {}", idx)), *b))
+      .collect::<Result<_, _>>()?;
+
+    Ok((commit_1, commit_2, result, scalar))
+  }
+
+  /// TODO: docs
+  pub fn synthesize<CS: ConstraintSystem<<E as Engine>::Base>>(
```
adr1anh, line 97: It's crucial that the secondary circuit sets `commit_1, commit_2, result, scalar` as public IO using `inputize`. Without these, the inputs can be entirely different from those derived from the primary circuit verifier.

adr1anh (commented):
First batch of comments

src/cyclefold/gadgets.rs
```diff
@@ -0,0 +1,586 @@
...
+  },
+  traits::{commitment::CommitmentTrait, Engine, ROCircuitTrait},
+};
+
+use bellpepper_core::{ConstraintSystem, SynthesisError};
+pub struct AllocatedFoldingData<E: Engine> {
+  pub U: AllocatedRelaxedR1CSInstance<E>,
+  pub u: AllocatedR1CSInstance<E>,
+  pub T: AllocatedPoint<E>,
+}
```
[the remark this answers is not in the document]
mpenciak: This `FoldingData` is made up of the "correct" field instances and points. It could even be used in the original Nova/SuperNova augmented circuit if we wanted. I added it here because it reduces a lot of code duplication from the data in the augmented circuit coming from the multiple invocations of the Cyclefold circuit.

src/cyclefold/gadgets.rs
```diff
@@ -0,0 +1,586 @@
...
+      cs.namespace(|| "U"),
+      inst.map(|x| &x.U),
+      limb_width,
+      n_limbs,
+    )?;
+
+    let u = AllocatedR1CSInstance::alloc(cs.namespace(|| "u"), inst.map(|x| &x.u))?;
+
+    let T = AllocatedPoint::alloc(cs.namespace(|| "T"), inst.map(|x| x.T.to_coordinates()))?;
+    T.check_on_curve(cs.namespace(|| "T on curve"))?;
```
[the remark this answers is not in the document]
mpenciak: This check is actually for the "natively cheap" allocated points. It matches the on curve checks that are done in the Nova/SuperNova `alloc_witness` functions.

src/cyclefold/circuit.rs
mpenciak: With the changes that @adr1anh proposes I don't actually need to check equality for `AllocatedPoint`s anymore.

wyattbenno777:
What would you say the best test against this would be for Nova vs ParaNova?

test_ivc_nontrivial_with vs test_trivial_cyclefold_prove_verify_with?

mpenciak:
> What would you say the best test against this would be for Nova vs ParaNova?
> 
> test_ivc_nontrivial_with vs test_trivial_cyclefold_prove_verify_with?

I would say those two tests are the most comparable.

But a word of warning: The current folding circuit is not in its final state. I'm nearing the end of a re-work for the way the public IO for the CycleFold circuit is dealt with. Right now the primary folding circuit has ~80k constraints, which is largely coming from the fact that the CycleFold IO has arity 16 (5 per commitment + 1 scalar). This will make the CycleFold circuit moderately larger (up to ~2k constraints), but significantly reduce the size of the folding circuit (numbers TBD soon).

Hopefully in the next day or so these updates will be pushed and there can be a more fair comparison.

mpenciak:
> What would you say the best test against this would be for Nova vs ParaNova?
> 
> test_ivc_nontrivial_with vs test_trivial_cyclefold_prove_verify_with?

Just a quick update: I just finished the IO refactor, so the comparison should be more fair now.

src/gadgets/nonnative/util.rs
```diff
@@ -155,8 +155,45 @@ impl<Scalar: PrimeField> Num<Scalar> {
     Ok(())
   }
 
-  /// Computes the natural number represented by an array of bits.
-  /// Checks if the natural number equals `self`
+  // TOOD: Can probably delete this
```
huitseeker, line 158: Let's.

src/gadgets/r1cs.rs
```diff
@@ -229,7 +229,7 @@ impl<E: Engine, const N: usize> AllocatedRelaxedR1CSInstance<E, N> {
     n_limbs: usize,
   ) -> Result<Self, SynthesisError> {
     // Compute r:
-    let mut ro = E::ROCircuit::new(ro_consts, NUM_FE_FOR_RO);
+    let mut ro = E::ROCircuit::new(ro_consts, 7 + N);
```
huitseeker, line 232: Please introduce another constant and avoid the magic number.

src/nifs.rs
```diff
@@ -48,9 +45,22 @@ impl<E: Engine> NIFS<E> {
...
+  > {
+    // Check `U1` and `U2` have the same arity
+    let io_arity = U1.X.len();
+    if io_arity != U2.X.len() {
+      return Err(NovaError::InvalidInputLength);
+    }
+
     // initialize a new RO
-    let mut ro = E::RO::new(ro_consts.clone(), NUM_FE_FOR_RO);
+    let mut ro = E::RO::new(ro_consts.clone(), 7 + io_arity);
```
huitseeker, line 63: Please introduce a new constant and avoid the magic number.

src/circuit.rs
```diff
@@ -459,8 +459,8 @@ mod tests {
       &params2,
       ro_consts1,
       ro_consts2,
-      &expect!["9817"],
```
huitseeker, line 462: 1. what's the logic behind the general nova circuit's constraint count changing?
2. can you please also update NUM_CONS_VERIFIER_CIRCUIT_PRIMARY ?

src/supernova/circuit.rs
```diff
@@ -836,8 +836,8 @@ mod tests {
       &params2,
       ro_consts1,
       ro_consts2,
-      &expect!["9836"],
-      &expect!["10384"],
+      &expect!["9844"],
+      &expect!["10392"],
```
huitseeker, line 840: It seems this needs a rebase, no? We're already at 9844 for SuperNova
https://github.com/lurk-lab/arecibo/blob/dev/benches/common/supernova/mod.rs#L31

src/circuit.rs
mpenciak: Oh this is funky, I must have messed up the last rebase. These constraint counts shouldn't change, I can fix it real quick.

src/gadgets/r1cs.rs
mpenciak: I noticed in trying to fix this that our use of `NUM_FE_WITHOUT_IO_FOR_CRHF` has drifted away from its original purpose, so I'm going to try to fix up the use of constants beyond just adding this one.

src/cyclefold/circuit.rs
```diff
@@ -0,0 +1,319 @@
...
+  },
+  traits::{commitment::CommitmentTrait, Engine},
+  Commitment,
+};
+
+/// TODO: docs
+pub struct CyclefoldCircuit<E: Engine> {
+  commit_1: Option<Commitment<E>>,
+  commit_2: Option<Commitment<E>>,
+  scalar: Option<[bool; NUM_CHALLENGE_BITS]>,
```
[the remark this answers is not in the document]
huitseeker: Edit: issue is #369

huitseeker (approved):
I've created https://github.com/lurk-lab/arecibo/issues/375 for the code duplication. THis looks great, thank you so much!

mpenciak:
> I've created #375 for the code duplication. THis looks great, thank you so much!

Thanks!

PR #276, opened 2024-01-20
Nova forward ports

huitseeker:
This ports the following upstream PRs:
- https://github.com/microsoft/Nova/pull/286 (no-op)
- https://github.com/microsoft/Nova/pull/289 (easy)
- https://github.com/microsoft/Nova/pull/290 (trivial)
- https://github.com/microsoft/Nova/pull/291 (non-trivial)
- https://github.com/microsoft/Nova/pull/294 (easy)

The next-to-last one should impact our upstream Supernova PR in minor ways, please review.

### DIFF
--- a/examples/and.rs
+++ b/examples/and.rs
@@ -1,0 +1,338 @@
+//! This example executes a batch of 64-bit AND operations.
+//! It performs the AND operation by first decomposing the operands into bits and then performing the operation bit-by-bit.
+//! We execute a configurable number of AND operations per step of Nova's recursion.
+use arecibo::{
+  provider::{Bn256EngineKZG, GrumpkinEngine},
+  traits::{
+    circuit::{StepCircuit, TrivialCircuit},
+    snark::RelaxedR1CSSNARKTrait,
+    Engine, Group,
+  },
+  CompressedSNARK, PublicParams, RecursiveSNARK,
+};
+use bellpepper_core::{
+  boolean::AllocatedBit, num::AllocatedNum, ConstraintSystem, LinearCombination, SynthesisError,
+};
+use core::marker::PhantomData;
+use ff::Field;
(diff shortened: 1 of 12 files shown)


mpenciak (approved):

adr1anh:
Since SuperNova calls the existing NIFS fold function, we shouldn't have to make any modification to the augmentation circuit, and should be getting these benefits "for free"

huitseeker:
> Since SuperNova calls the existing NIFS fold function, we shouldn't have to make any modification to the augmentation circuit, and should be getting these benefits "for free"

I'm not sure whether your remark is in the context of analyzing the present PR, or about the impact on our Supernova PR. I'll answer each

## In the context of the present PR:

In theory, yes, the changes from Nova should be one trivial rebase from fitting into our repo. In practice, we have a separate set of non-allocating functions which are not upstream, that are mostly exemplified by https://github.com/microsoft/Nova/pull/277, and so the code locations modified by Nova upstream need to have equivalent code locations modified in these non-alloc functions (e.g. our nifs/[prove_mut](https://github.com/lurk-lab/arecibo/pull/276/files#diff-554a4aa7316926ea9a91db667f1de90705985e19b630fc591e93e66e737eeb3eL87-L107))

Further, the changes of the upstream PR have impact on some SuperNova tests, such as constraint counts in the circuit. These SuperNova tests are simply not in the upstream repo, so we just have to compute (through expect_test) the new values.

## In the context of impacting the upstream PR:

The upstream Supernova PR *introduces* tests that are impacted by the RO changes of the next-to-last commit, e.g. constraint counts for the SuperNova circuit, which are not upstream. The introduced material will need to be changed upon rebasing the PR.

adr1anh (approved):

PR #279, opened 2024-01-24
Enable `grumpkin-msm` CUDA feature

samuelburnham:
Adapts to https://github.com/lurk-lab/grumpkin-msm/pull/9, enabling GPU acceleration when the `cuda` feature is enabled


huitseeker (approved):
:shipit:

PR #286, opened 2024-01-30
chore: remove subsumed test (easy)

huitseeker:
See #222, the commit_open_prove_verify test is subsumed by the subsequent unit test. @tchataigner was right from the start. 🤦


mpenciak (approved):
Just a comment to clarify this for myself (and others stumbling upon this issue):

This test is subsumed by `test_multiple_polynomial_size()` because both tests are testing the same transcript equality (see in the testing utils [here](https://github.com/tchataigner/arecibo/blob/452f95c9ab4141bfc30aaf4499aae46e6c475e06/src/provider/util/mod.rs#L110))

PR #287, opened 2024-01-31
Remove superfluous instances of derive(Clone) (easy)

huitseeker:
This should help us be more rigorous with cloning. Tested to not break Lurk in the (currently unused) scenario where we'd use pre-processing variants of compressing SNARKs. There will probably be another, slightly more extensive cleanup after #285 merges (as that PR has the downstream effect of shrinking Lurk's requirements on Arecibo, see https://github.com/lurk-lab/lurk-rs/pull/1085).

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -3,7 +3,7 @@ use core::fmt::Debug;
 use thiserror::Error;
 
 /// Errors returned by Nova
-#[derive(Clone, Debug, Eq, PartialEq, Error)]
+#[derive(Debug, Eq, PartialEq, Error)]
 #[non_exhaustive]
 pub enum NovaError {
   /// returned if the supplied row or col in (row,col,val) tuple is out of range
@@ -72,7 +72,7 @@ pub enum NovaError {
 }
 
 /// Errors specific to the Polynomial commitment scheme
-#[derive(Clone, Debug, Eq, PartialEq, Error)]
+#[derive(Debug, Eq, PartialEq, Error)]
 pub enum PCSError {
   /// returned when an invalid inner product argument is provided
   #[error("InvalidIPA")]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -818,7 +818,7 @@ where
 }
 
 /// A SNARK that proves the knowledge of a valid `RecursiveSNARK`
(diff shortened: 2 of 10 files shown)


src/gadgets/nonnative/util.rs
```diff
@@ -18,7 +18,7 @@ pub struct Bit<Scalar: PrimeField> {
   pub value: Option<bool>,
 }
 
-#[derive(Clone)]
+#[derive()]
```
adr1anh: Can we just remove the line?

src/provider/non_hiding_zeromorph.rs
```diff
@@ -90,7 +90,7 @@ fn trim<E: Engine>(
 }
 
 /// Commitments
-#[derive(Debug, Clone, Eq, PartialEq, Default, Serialize, Deserialize)]
+#[derive(Debug, Eq, PartialEq, Default, Serialize, Deserialize)]
```
adr1anh, line 93: Given that `E::G1Affine` implements `Copy` I think we can actually keep this one.

src/provider/non_hiding_zeromorph.rs
```diff
@@ -109,7 +109,7 @@ impl<E: Engine> From<ZMCommitment<E>> for UVKZGCommitment<E> {
 }
 
 /// Polynomial Evaluation
-#[derive(Debug, Clone, Eq, PartialEq, Default)]
+#[derive(Debug, Eq, PartialEq, Default)]
```
adr1anh, line 112: Given that `E::Fq` implements `Copy` I think we can actually keep this one.

src/gadgets/ecc.rs
```diff
@@ -590,7 +590,7 @@ where
   }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
```
adr1anh, line 593: This one seems fine as most `Allocated*` types support `Clone`.

adr1anh (commented):
I fully support removing the ability for cloning most of these structs. However, I think it's fine to leave it for small types where the underlying value already implements `Copy`.

mpenciak (approved):

mpenciak (approved):

PR #295, opened 2024-02-02
Test effect of cuda feature

huitseeker:
#294 from a main branch


huitseeker:
!gpu-benchmark

huitseeker:
The issue in #281 is not a regression but is a reproducible perf issue

PR #305, opened 2024-02-08
Rework batches in ppsnark

huitseeker:
## TL;DR

This is streamlining of the batched snark/ppsnark made on the way to resolving the perf difference found in #281 

Differential bench (with `--features "cuda"`) for this PR reveals high variance 🤷 

### CompressedSNARK-NIVC-1

|                              | `ref=13ff737`             | `ref=7a5d7bf`                     |
|:-----------------------------|:--------------------------|:--------------------------------- |
| **`Prove-NumCons-6540`**     | `378.03 ms` (✅ **1.00x**) | `384.75 ms` (✅ **1.02x slower**)  |
| **`Verify-NumCons-6540`**    | `29.71 ms` (✅ **1.00x**)  | `29.00 ms` (✅ **1.02x faster**)   |
| **`Prove-NumCons-1038732`**  | `9.42 s` (✅ **1.00x**)    | `9.42 s` (✅ **1.00x slower**)     |
| **`Verify-NumCons-1038732`** | `132.44 ms` (✅ **1.00x**) | `115.03 ms` (✅ **1.15x faster**)  |

### CompressedSNARK-NIVC-2

|                              | `ref=13ff737`             | `ref=7a5d7bf`                     |
|:-----------------------------|:--------------------------|:--------------------------------- |
| **`Prove-NumCons-6540`**     | `384.48 ms` (✅ **1.00x**) | `388.90 ms` (✅ **1.01x slower**)  |
| **`Verify-NumCons-6540`**    | `31.16 ms` (✅ **1.00x**)  | `29.68 ms` (✅ **1.05x faster**)   |
| **`Prove-NumCons-1038732`**  | `9.73 s` (✅ **1.00x**)    | `10.03 s` (✅ **1.03x slower**)    |
| **`Verify-NumCons-1038732`** | `182.09 ms` (✅ **1.00x**) | `149.88 ms` (✅ **1.21x faster**)  |

### CompressedSNARK-NIVC-Commitments-2

|                              | `ref=13ff737`             | `ref=7a5d7bf`                     |
|:-----------------------------|:--------------------------|:--------------------------------- |
| **`Prove-NumCons-6540`**     | `9.30 s` (✅ **1.00x**)    | `9.41 s` (✅ **1.01x slower**)     |
| **`Verify-NumCons-6540`**    | `57.65 ms` (✅ **1.00x**)  | `57.70 ms` (✅ **1.00x slower**)   |
| **`Prove-NumCons-1038732`**  | `70.85 s` (✅ **1.00x**)   | `70.12 s` (✅ **1.01x faster**)    |
| **`Verify-NumCons-1038732`** | `299.04 ms` (✅ **1.00x**) | `300.82 ms` (✅ **1.01x slower**)  |

### DIFF
--- a/src/spartan/batched.rs
+++ b/src/spartan/batched.rs
@@ -403,14 +403,14 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> BatchedRelaxedR1CSSNARKTrait<E>
     let num_rounds_y_max = *num_rounds_y.iter().max().unwrap();
 
     // Define τ polynomials of the appropriate size for each instance
-    let polys_tau = {
-      let tau = transcript.squeeze(b"t")?;
+    let tau = transcript.squeeze(b"t")?;
+    let all_taus = PowPolynomial::squares(&tau, num_rounds_x_max);
 
-      num_rounds_x
-        .iter()
-        .map(|&num_rounds| PowPolynomial::new(&tau, num_rounds))
-        .collect::<Vec<_>>()
-    };
+    let polys_tau = num_rounds_x
+      .iter()
+      .map(|&num_rounds_x| PowPolynomial::evals_with_powers(&all_taus, num_rounds_x))
+      .map(MultilinearPolynomial::new)
+      .collect::<Vec<_>>();
 
     // Sample challenge for random linear-combination of outer claims
     let outer_r = transcript.squeeze(b"out_r")?;
--- a/src/spartan/batched.rs
+++ b/src/spartan/batched.rs
@@ -156,9 +156,7 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> BatchedRelaxedR1CSSNARKTrait<E>
       let num_instances_field = E::Scalar::from(num_instances as u64);
       transcript.absorb(b"n", &num_instances_field);
     }
-    for u in U.iter() {
-      transcript.absorb(b"U", u);
-    }
+    transcript.absorb(b"U", &U);
 
     let (polys_W, polys_E): (Vec<_>, Vec<_>) = W.into_iter().map(|w| (w.W, w.E)).unzip();
 
@@ -395,9 +393,7 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> BatchedRelaxedR1CSSNARKTrait<E>
       let num_instances_field = E::Scalar::from(num_instances as u64);
       transcript.absorb(b"n", &num_instances_field);
     }
-    for u in U.iter() {
-      transcript.absorb(b"U", u);
-    }
+    transcript.absorb(b"U", &U);
 
     let num_instances = U.len();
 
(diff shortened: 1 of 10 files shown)


storojs72 (commented):
The Rust specific simplifications are indeed awesome!

I'm not sure if I understand why do you put benches results comparison for 
```
ref=13ff737 / ref=7a5d7bf
```

as [mentioned](https://github.com/lurk-lab/arecibo/issues/281) performance regression was detected for https://github.com/lurk-lab/arecibo/commit/5da330e6ed2c39679f2cf591edf25543d7a0399b, so it is natural to compare:
```
ref=5da330e (regression detected) / ref=aed2791 (last commit of this PR)
```
isn't it?

huitseeker:
@storojs72 I have added more information to #281 to give context about the performance issue of Supernova proofs raised there. I would note that this is technically *not* a regression: all our benchmarks were on the GPU before, and #279 (joined with updates on the grumpkin-msm repo) is the first PR where we have seen the performance of SuperNova proofs w/o the GPU.

Moreover, this PR is *not* aimed at solving the issue of Supernova proofs. It's simplifications I performed *on the way* towards addressing that issue. They should stand on their own. As you can see they are not addressing changes related to GPU operations.

src/spartan/batched.rs
```diff
@@ -134,17 +134,7 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> BatchedRelaxedR1CSSNARKTrait<E>
...
-      .map(|s| {
-        let s = s.pad();
-        if s.is_regular_shape() {
-          Ok(s)
-        } else {
-          Err(NovaError::InternalError)
-        }
-      })
-      .collect::<Result<Vec<_>, _>>()?;
+    let S = S.iter().map(|s| s.pad()).collect::<Vec<_>>();
```
storojs72, line 137: I think here we skip `is_regular_shape` checks. Isn't this a problem?

src/spartan/ppsnark.rs
```diff
@@ -604,14 +604,12 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> RelaxedR1CSSNARKTrait<E> for Relax
...
-        let val = pk
-          .S_repr
-          .val_A
-          .par_iter()
-          .zip_eq(pk.S_repr.val_B.par_iter())
-          .zip_eq(pk.S_repr.val_C.par_iter())
-          .map(|((v_a, v_b), v_c)| *v_a + c * *v_b + c * c * *v_c)
-          .collect::<Vec<E::Scalar>>();
+        let val = zip_with!(
+          par_iter,
```
storojs72, line 608: Out of curious - how does it work with parallel iterator if indices of `valA`, `valB`, `valC` should correspond? (if I understand this correctly)

src/spartan/batched.rs
huitseeker: The goal of the `pad` function is to return things with regular shape. It already short-circuits when the input has that regular shape.

src/spartan/ppsnark.rs
huitseeker: The parallel iterator one gets from vectors implements [IndexedParallelIterator](https://docs.rs/rayon/latest/rayon/iter/trait.IndexedParallelIterator.html), which captures this "matching indices" guarantee.

storojs72 (approved):

PR #308, opened 2024-02-10
ci: Use configurable `!benchmark` PR comment

samuelburnham:
# Configurable `!benchmark` PR comment
Uses the reusable workflow from https://github.com/lurk-lab/ci-workflows/pull/38. This replaces the existing `!gpu-benchmark` command with 
`!benchmark`, as shown in the following examples:
```
!benchmark
```
This will run the default benchmark settings, specified in the workflow file of this PR.
Alternatively: 
```
!benchmark --bench supernova-ci --bench pcs --features cuda
ARECIBO_BENCH_NUM_CONS=16384,1048576
```
This will run with each of the input benchmarks, features, and env vars.

> [!IMPORTANT]
> - If you want to benchmark with GPU acceleration enabled, you *must* specify `--features cuda` in the comment body (see above example). Otherwise, the benchmarks will still run on the same GPU-equipped benchmark machine, but will run using the CPU only.

## Details
- The previous `!gpu-benchmark` syntax will no longer work. The command must start with `!benchmark`.
- Each benchmark specified with `--bench` will spawn a new CI job, which means they will each output a separate comment.

The default env vars specified in this PR are 
```
ARECIBO_BENCH_OUTPUT=pr-comment
ARECIBO_BENCH_NUM_CONS=16384,524288
```
If identically named env vars are specified in the `!benchmark` comment, they will overwrite the above defaults.

See also https://github.com/lurk-lab/lurk-rs/pull/1114

### DIFF
--- a/.github/workflows/bench_pr_comment.yml
+++ b/.github/workflows/bench_pr_comment.yml
@@ -1,4 +1,4 @@
-# Creates a PR benchmark comment with a comparison to main
+# Creates a PR benchmark comment with a comparison to the base branch
 name: Benchmark pull requests
 on:
   issue_comment:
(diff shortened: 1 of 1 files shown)


huitseeker (approved):
Thanks a lot!

PR #310, opened 2024-02-10
ci: Add nightly typo checker workflow

samuelburnham:
Uses the typo checker workflow added in https://github.com/lurk-lab/ci-workflows/pull/29, which opens a PR with typo fixes and comments unsolvable typos in the PR description.

I tested the output of `typos --write-changes` locally and it seems to work quite well, no false positives that I could see. If merged, we should get a nice-looking PR on the next morning.


huitseeker (approved):
🚀

PR #315, opened 2024-02-11
Refactor Polyeval{Instance, Witness}

huitseeker:
After #305 (which it includes, so go review there plz)

Refactors Polynomial evaluation instances / witnesses to express the specialized functions in terms of the general ones.

For review: the refactoring of `PolyEvalInstance::batch_diff_size`, which is strict about the size of its arguments and inspects it ,to be the implementation of `PolyEvalInstance::batch`, has revealed the later call sites are *not* rigorous about the size of their arguments, nor do they inspect them.

I've patched call-sites to be compliant with the rigor of `batch_diff_size`, but there's another "looser" variant of this PR that would eliminate `x` from the `PolyEvalInstance`.

### DIFF
--- a/src/spartan/mod.rs
+++ b/src/spartan/mod.rs
@@ -24,6 +24,7 @@ use crate::{
 use ff::Field;
 use itertools::Itertools as _;
 use polys::multilinear::SparsePolynomial;
+
 use rayon::{iter::IntoParallelRefIterator, prelude::*};
 
 // Creates a vector of the first `n` powers of `s`.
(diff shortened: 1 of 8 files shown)


src/spartan/ppsnark.rs
```diff
@@ -580,7 +580,12 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> RelaxedR1CSSNARKTrait<E> for Relax
     let poly_vec = vec![&Az, &Bz, &Cz];
     let c = transcript.squeeze(b"c")?;
     let w: PolyEvalWitness<E> = PolyEvalWitness::batch(&poly_vec, &c);
-    let u: PolyEvalInstance<E> = PolyEvalInstance::batch(&comm_vec, &tau_coords, &eval_vec, &c);
+    let u: PolyEvalInstance<E> = PolyEvalInstance::batch(
+      &comm_vec,
+      tau_coords[..comm_vec.len()].to_vec(),
```
adr1anh, line 585: `tau_coords` should be copied as is, its size does not depend on the number of commitments.

src/spartan/mod.rs
```diff
@@ -225,3 +204,124 @@ fn compute_eval_table_sparse<E: Engine>(
...
+          // perform vector addition to combine the weighted vectors
+          acc.into_iter().zip_eq(v).map(|(x, y)| x + y).collect()
+        },
+      );
+
+      Self { p }
+    }
+  }
+
+  impl<E: Engine> PolyEvalInstance<E> {
```
storojs72, line 243: Now with generic `PolyEvalInstance` we can derive various implementations of it, right? Do we use that already (or have plans to create and use) and if so where concretely (more exactly - do we have / need alternative implementation which is different comparing to the previous [reference](https://github.com/lurk-lab/Nova/blob/af601ed0513a2df6321a2d88293ca3260e71ef84/src/spartan/mod.rs#L87) one implemented at Solidity side of things)?

huitseeker: `PolyEvalInstance` has been generic since the beginning, and the type itself has not changed:
https://github.com/microsoft/Nova/blob/37871d04ba04771bf19c51e5b336e45fdb86c7f7/src/spartan/mod.rs#L117-L121
I'm not sure I understand the question?

storojs72: Right, the question is actually - does new PolyEvalInstanse preserve compatibility with older one and where alt_batch is planning to be used (since as far as I see it is absent in upstream)

adr1anh (approved):

storojs72 (approved):

PR #322, opened 2024-02-14
Make `AllocatedR1CSInstance` and `AllocatedRelaxedR1CSInstance` generic in the size of IO

### LINKED ISSUE #321: Make `AllocatedR1CSInstance` and `AllocatedRelaxedR1CSInstance`  generic in size of IO
Right now in the codebase [`AllocatedR1CSInstance`](https://github.com/lurk-lab/arecibo/blob/4cc09c82fff8e4aa5821f5ce58a217989d105c67/src/gadgets/r1cs.rs#L25-L28) and [`AllocatedRelaxedR1CSInstance`](https://github.com/lurk-lab/arecibo/blob/4cc09c82fff8e4aa5821f5ce58a217989d105c67/src/gadgets/r1cs.rs#L61-L67) have their IO hardcoded to be of length 2, with the structs having the parameters `X0` and `X1`. This is at odds with `R1CSInstance` and `RelaxedR1CSInstance` which can in principle have any sized IO.

The IO of length 2 works for the Nova and SuperNova folding usecases, but in CycleFold we need allocated R1CS Instances with different numbers of IO.

mpenciak:
This PR resolves #321.

The solution adds a const generic parameter to the two structs, and a new constant that is used in both the Nova and SuperNova augmented circuits where the instances are known to have IO of length 2.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -5,7 +5,7 @@ pub(crate) const BN_LIMB_WIDTH: usize = 64;
 pub(crate) const BN_N_LIMBS: usize = 4;
 pub(crate) const NUM_FE_WITHOUT_IO_FOR_CRHF: usize = 17;
 pub(crate) const NUM_FE_FOR_RO: usize = 9;
-pub(crate) const NUM_IO_IN_NOVA: usize = 2;
+pub(crate) const NIO_NOVA_FOLD: usize = 2;
 
 /// Bit size of Nova field element hashes
 pub const NUM_HASH_BITS: usize = 250;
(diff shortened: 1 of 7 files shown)


src/constants.rs
```diff
@@ -5,6 +5,7 @@ pub(crate) const BN_LIMB_WIDTH: usize = 64;
 pub(crate) const BN_N_LIMBS: usize = 4;
 pub(crate) const NUM_FE_WITHOUT_IO_FOR_CRHF: usize = 17;
 pub(crate) const NUM_FE_FOR_RO: usize = 9;
+pub(crate) const NUM_IO_IN_NOVA: usize = 2;
```
huitseeker, line 8: We're using this constant in the Supernova module: this is going to confuse the heck out of people. Would something more specific help? maybe `NIO_NOVA_FOLD`?

huitseeker (commented):
From a low level PoV, the code is sometimes a bit complex, but can be simplified, see inline comments.

From a high-level PoV, this gives use generic code in gadgets, that can be used in a scheme-specific manner outside that module, which I think fits the bill: awesome!

mpenciak:
I think I cleaned up some of the nasty stuff I was doing with iterators, it should be a bit better now!

huitseeker (approved):
Thanks!

PR #327, opened 2024-02-15
Eliminate hard-coded `BN_N_LIMBS` and `BN_LIMB_WIDTH` dependency.

mpenciak:
In an attempt to see what would happen to the number of constraints if we changed the parameters `BN_N_LIMBS` and `BN_LIMB_WIDTH`, I found a couple places where the arithmetic inadvertently was dependent on the values being hardcoded as 4 and 64 respectively.

I believe this PR makes the necessary changes to make the codebase general enough to allow for varying these parameters. The expect tests expectedly (:drum: haha) fail when changing these two constants, but everything else passes.

### DIFF
--- a/src/gadgets/nonnative/bignat.rs
+++ b/src/gadgets/nonnative/bignat.rs
@@ -444,7 +444,11 @@ impl<Scalar: PrimeField> BigNat<Scalar> {
     let carry_bits = (((max_word.to_f64().unwrap() * 2.0).log2() - self.params.limb_width as f64)
       .ceil()
       + 0.1) as usize;
-    let limbs_per_group = (Scalar::CAPACITY as usize - carry_bits) / self.params.limb_width;
+    let limbs_per_group = max(
+      (Scalar::CAPACITY as usize - carry_bits) / self.params.limb_width,
+      1,
+    );
+
     let self_grouped = self.group_limbs(limbs_per_group);
     let other_grouped = other.group_limbs(limbs_per_group);
     self_grouped.equal_when_carried(cs.namespace(|| "grouped"), &other_grouped)
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -3,7 +3,7 @@
 pub(crate) const NUM_CHALLENGE_BITS: usize = 128;
 pub(crate) const BN_LIMB_WIDTH: usize = 64;
 pub(crate) const BN_N_LIMBS: usize = 4;
-pub(crate) const NUM_FE_WITHOUT_IO_FOR_CRHF: usize = 17;
+pub(crate) const NUM_FE_WITHOUT_IO_FOR_CRHF: usize = 9 + NIO_NOVA_FOLD * BN_N_LIMBS;
 pub(crate) const NUM_FE_FOR_RO: usize = 9;
 pub(crate) const NIO_NOVA_FOLD: usize = 2;
 


src/gadgets/nonnative/bignat.rs
```diff
@@ -444,7 +444,11 @@ impl<Scalar: PrimeField> BigNat<Scalar> {
     let carry_bits = (((max_word.to_f64().unwrap() * 2.0).log2() - self.params.limb_width as f64)
       .ceil()
       + 0.1) as usize;
```
adr1anh, line 446: This seems like it could lead to unexpected results in some situations due to a potential loss of precision with floats.

src/gadgets/nonnative/bignat.rs
```diff
@@ -444,7 +444,11 @@ impl<Scalar: PrimeField> BigNat<Scalar> {
...
-    let limbs_per_group = (Scalar::CAPACITY as usize - carry_bits) / self.params.limb_width;
+    let limbs_per_group = max(
+      (Scalar::CAPACITY as usize - carry_bits) / self.params.limb_width,
+      1,
+    );
+
```
adr1anh, line 451: I am not sure of the exact arithmetic, but would 
```rust
let limbs_per_group = (Scalar::CAPACITY as usize - carry_bits + (self.params.limb_width-1)) / self.params.limb_width
```
may be closer to the intended behavior.

adr1anh:
The change looks good to me, though I haven't looked into this gadget enough to understand if this is enough to support arbitrary configurations. Ideally, we should replace this with [bellpepper-gadgets/emulated](https://github.com/lurk-lab/bellpepper-gadgets/tree/main/crates/emulated).

src/gadgets/nonnative/bignat.rs
mpenciak: My intention with the change was to ensure that the number of limbs per group is at least 1 without modifying the original behavior of the circuit. This came up in the case with BN width of 128 and limb number 2 where the original calculation results in `limbs_per_group = 0`, leading to a divide by zero panic when calculating `n_limbs` in `group_limbs(...)`. 

With the change you propose the `limbs_per_group` matches for the 128 x 2 configuration, but also has the effect of changing the value of `limbs_per_group` in other cases as well. In particular in the 64 x 4 configuration this leads to 3 `limbs_per_group` compared to the original 2. Interestingly this leads to a slightly smaller circuit, but it also causes tons of panics in tests checking IVC proof/verification, so I think it causes some issues with the original logic.

adr1anh: Hmm, I think I didn't correctly understand the intention of this value. My suggestions was essentially performing a `ceil` of the `(Scalar::CAPACITY as usize - carry_bits) / self.params.limb_width` value in a safe way, but I don't think that was actually intended here .

huitseeker:
@adr1anh @mpenciak 
Hi folks, could we get a decision on this PR with the code base we have **today**? Surely it seems a 6-line change is a thing we can make a decision (positive or negative) today, even if the ultimate goal of moving to bellpepper-emulated seems optimal tomorrow?

adr1anh:
The change does seem reasonable to me, and can be backported to Nova too, even if we move to `bellpepper-emulated` later.

Issue #270, opened 2024-01-17
Load KZG setup parameters from file


huitseeker:
This is a copy of https://github.com/microsoft/Nova/issues/270

With #145, and #172, #206, #207, #231, #242, #260, we've added MLKZG & Zeromorph support, which depend on a universal trusted setup. We have a weirder shape of the Universal trusted setup struct :
https://github.com/lurk-lab/arecibo/blob/aaa13396cdb8776b11a68e7bccce4654399e52c8/src/provider/non_hiding_kzg.rs#L24-L33

.. when compared to upstream nova, to accommodate the shape required by Zeromorph. It's expected we would simply make the lenght of the `powers_of_h` field equal to 2 when using a garden-variety KZG trusted setup, which would make it work just as well (see the [upstream struct for comparison](https://github.com/microsoft/Nova/blob/8244d00a7b3bf5fc44e8cf76c3a0d67cfdb17394/src/provider/mlkzg.rs#L34-L40)).

Note the setup is "as is" in production: if deserialized and then found to be lacking (e.g. at the call of a trim function that generates proof-specific `ProverKey`, `VerifierKey` and finds the number of bases insufficient), there is no remedy but an irrecoverable error.

> [!NOTE]
> We can use our knowledge of abomonation/deabomonation, along with serialziation, in order to make the loading of these parameters faster.

> [!TIP]
> We could also perform the same thing for Pedersen public parameters, since they are generated deterministically, and a cached file would be a quality of life improvement.

PR #335, opened 2024-02-20
Add API for generic data directory

winston-h-zhang:
This PR implements a minimal generic data directory for `arecibo` to read/write from. It is similar the the data directory used in Lurk. Besides the immediate need to have some way of managing witness/cross-term vectors for our MSM data analysis, there has been a need for this type of infrastructure for a while. 

See, for example, [https://github.com/microsoft/Nova/issues/270, which mentions the need ](https://github.com/lurk-lab/arecibo/issues/270), which can build upon this PR. There is also https://github.com/lurk-lab/lurk-rs/issues/1086, which may need to move some of Lurk's caching logic into `arecibo`.

This PR doesn't intend to address all the issues of these PRs, but it would be nice to have the first increment of work streamlined into the codebase, so that we can slowly build up to a robust solution.


wwared (changes requested):
Not commenting on whether this PR solves the issues around loading/writing data in the way we want to or not, but wanted to point out a few things regarding this code regardless of that:

* Since we call `set` or `get_or_init` from within the `prove_step` function (by calling `write_data()`), isn't it possible that `ARECIBO_CONFIG` could be initialized from different threads? (even possibly from different threads proving different, unrelated things) It might be better to use [OnceLock](https://doc.rust-lang.org/stable/std/sync/struct.OnceLock.html) instead if that's the case (`OnceCell` is not thread-safe) (it should not be a big deal to change because it may only block during initialization)
* The cfg overrides feel weird to me, shouldn't it be `cfg(target_arch = "wasm32")` instead of `cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))`?
* I feel like it'd be simpler if the `data` mod was present on every arch, but for the arches we disable it (wasm) just replace with dummy functions instead. This prevents the `#[cfg]` directives from polluting the callers of the data module
* Usability-wise, it would be good to support overriding the `~/.arecibo_data` path with an env var like `ARECIBO_DATA_DIR` or similar
* I know it's meant to be extended in the future, but it's weird how `write_data` is duplicated on the global, per-process `DataConfig` stored in `ARECIBO_CONFIG` as well as in the `RecursiveSNARKConfig` present in every `RecursiveSNARK`. I think for now it would be better to not have the `RecursiveSNARKConfig` at all and keep configuration (enabling/disabling the write data) on the `data` module, possibly controlled by other env vars

winston-h-zhang:
Thank you for all the comments! I've written a starting issue here: https://github.com/lurk-lab/arecibo/issues/371. It's more of a set of questions that I have for how to proceed, rather than what I think needs to be done, so I would be happy to discuss and create a plan for what's next.

PR #342, opened 2024-02-22
(slightly) Faster ML evaluations

### LINKED ISSUE #293: Cache MLE in repeated evaluations
We often evaluate multilinear extensions at the same point repeatedly, for instance:
https://github.com/lurk-lab/arecibo/blob/b168f2784af6825e88c5b2717907a0f278d3616e/src/spartan/snark.rs#L181-L182

In each case, the evaluation functions compute an MLE for the point, which is then used to compute the evaluation:
https://github.com/lurk-lab/arecibo/blob/b168f2784af6825e88c5b2717907a0f278d3616e/src/spartan/polys/multilinear.rs#L105-L128

See the repeat use of `EqPolynomial::evals_from_points(r)`

We could cache the MLE and reuse it in the second evaluations. There's several ways to approach this:
- explicitly and manually caching the MLE. This is useful because in the example above, we add insult to injury by explicitly computing the same MLE later: https://github.com/lurk-lab/arecibo/blob/b168f2784af6825e88c5b2717907a0f278d3616e/src/spartan/snark.rs#L194
- having an API for computing evaluations at the same point for several polynomials `evaluate(polys: &[MultilinearPolynomial<E>], r:[E::Scalar])`

huitseeker:
Slightly faster evaluations by caching chi coefficients between evaluations.

Fixes #293 .

### DIFF
--- a/src/spartan/batched.rs
+++ b/src/spartan/batched.rs
@@ -434,9 +434,15 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> BatchedRelaxedR1CSSNARKTrait<E>
       )
     }
 
+    let chis_r_x = r_x
+      .par_iter()
+      .map(|r_x| EqPolynomial::evals_from_points(r_x))
+      .collect::<Vec<_>>();
+
     // Evaluate τ(rₓ) for each instance
-    let evals_tau = zip_with!(iter, (polys_tau, r_x), |poly_tau, r_x| poly_tau
-      .evaluate(r_x));
+    let evals_tau = zip_with!(iter, (polys_tau, chis_r_x), |poly_tau, er_x| {
+      MultilinearPolynomial::evaluate_with_chis(poly_tau.evaluations(), er_x)
+    });
 
     // Compute expected claim for all instances ∑ᵢ rⁱ⋅τ(rₓ)⋅(Azᵢ⋅Bzᵢ − uᵢ⋅Czᵢ − Eᵢ)
     let claim_outer_final_expected = zip_with!(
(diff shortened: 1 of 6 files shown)


src/spartan/batched.rs
```diff
@@ -435,9 +435,10 @@ impl<E: Engine, EE: EvaluationEngineTrait<E>> BatchedRelaxedR1CSSNARKTrait<E>
       )
     }
 
+    let e_r_x = r_x.par_iter().map(|r_x| EqPolynomial::evals_from_points(&r_x)).collect::<Vec<_>>();
```
adr1anh, line 438: I got a tiny bit confused until I realized it was the `eq` polynomial. Could re rename it to `eq_r_x`?

adr1anh: Or even `chi_r_x`

src/spartan/polys/multilinear.rs
```diff
@@ -105,23 +105,20 @@ impl<Scalar: PrimeField> MultilinearPolynomial<Scalar> {
...
+
+  /// Evaluates the polynomial with the given evaluations and chi coefficients
+  pub fn evaluate_with_chis(Z: &[Scalar], chis: &[Scalar]) -> Scalar {
     zip_with!(
-      (
-        EqPolynomial::evals_from_points(r).into_par_iter(),
-        Z.par_iter()
-      ),
-      |a, b| a * b
+      (chis.into_par_iter(), Z.par_iter()),
```
adr1anh, line 120: The `into_` is not needed.

adr1anh (commented):
I just realized it's still a draft, but this has been annoying me for a while and happy to see this optimization!

adr1anh (approved):

PR #345, opened 2024-02-25
Batch inversion & normalization

huitseeker:
- Uses Montgomery-style batch inversion from ff:Field, rather than reimplementing in two places,
- Uses batch normalization for HyperKZG inversions.


adr1anh (approved):
Nice!

PR #346, opened 2024-02-25
No more assert on result

huitseeker:
An assertion failure on a Result cannot output the useful message of the error.

This removes the systematic pattern of asserting on result states, specially when it comes to tests / benches.
Prohibits them for good measure, using [clippy](https://rust-lang.github.io/rust-clippy/master/index.html#/assertions_on_result_states).


adr1anh (approved):
This would've helped a while ago, very happy to see this refactor!

PR #347, opened 2024-02-26
Public gadget utilities

tchataigner:
In `arecibo` we have a few gadget utilities that have been written. So far, they are only available to this crate. While working in `bellpepper-gadgets` or any other gadget repositories, a few of those utilities could come in handy.

There are already cases [where they have been copied](https://github.com/lurk-lab/bellpepper-gadgets/blob/94034f6dd90153dd446aa898ac49dcb1d9230c45/crates/emulated/src/field_element.rs#L559) so I believe that making them public should be alright.

### DIFF
--- a/src/gadgets/mod.rs
+++ b/src/gadgets/mod.rs
@@ -2,4 +2,4 @@
 pub mod ecc;
 pub(crate) mod nonnative;
 pub(crate) mod r1cs;
-pub(crate) mod utils;
+pub mod utils;


adr1anh:
I just wanted to link https://github.com/lurk-lab/bellpepper/issues/56 which was a discussion around unifying the different gadgets in the 3 crates (bellbepper-gadgets, lurk-rs, and arecibo). My personal view is that users should not be depending on arecibo for gadgets, since that is not the purpose of the crate. We should look into adding these to bellpepper-gadgets instead

tchataigner:
> I just wanted to link [lurk-lab/bellpepper#56](https://github.com/lurk-lab/bellpepper/issues/56) which was a discussion around unifying the different gadgets in the 3 crates (bellbepper-gadgets, lurk-rs, and arecibo). My personal view is that users should not be depending on arecibo for gadgets, since that is not the purpose of the crate. We should look into adding these to bellpepper-gadgets instead

Thanks! I wanted to get the discussion on the chat to continue on Github, what you say seems like a good alternative. I'll create an issue on `bellpepper-gadget` so that those utilities get ported there :)

PR #358, opened 2024-03-05
add small util for SuperNova testing (easy)

wyattbenno777:
added num_constraints_and_variables function

### DIFF
--- a/src/supernova/mod.rs
+++ b/src/supernova/mod.rs
@@ -435,6 +435,13 @@ where
     )
   }
 
+  pub fn num_constraints_and_variables_secondary(&self) -> (usize, usize) {
+    (
+      self.circuit_shape_secondary.r1cs_shape.num_cons,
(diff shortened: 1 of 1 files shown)


huitseeker (commented):
This looks great, if we look at the Nova analogous part, we also return the same values for the secondary:

https://github.com/lurk-lab/arecibo/blob/07f571e11f346ef51d4c4ce3cf14ab0360982a30/src/lib.rs#L326-L340

Now, I 💯 agree with you that the circuit counts at a particular circuit index make more sense, so thanks, this addition is valuable. Would you be up for adding another function returning constraint and variable count for the secondary as well while we're at it?

wyattbenno777:
> circuit_shape_secondary

Added!

huitseeker (approved):

wwared (approved):

arthurpaulino (approved):

porcuquine (approved):

PR #359, opened 2024-03-07
Refinement in IPA templating

storojs72:
This PR helps in partial solving of https://github.com/lurk-lab/solidity-verifier/issues/52. 

It adds more accurate representing of `CompressedCommitment` value when using it in Solidity templating. "More accurate" means that Grumpkin point compression is also taken to consideration.

To produce compatible unit test for IPA in Solidity, requires merging https://github.com/lurk-lab/solidity-verifier/pull/63


tchataigner (approved):
LGTM!

PR #361, opened 2024-03-11
[WIP] Introducing PCS codebase as a separate module

### LINKED ISSUE #304: PCS codebase refactoring
Our HyperKZG implementation as well as Zeromorph significantly rely on plain KZG10 implementation. We can utilise common KZG10 stuff, avoid code duplication and improve organisation.

storojs72:
Partially fixes https://github.com/lurk-lab/arecibo/issues/304.

Reorganises PCS codebase into a separate module. 

Moves vanilla KZG10 utilities from ZM. 

Still work in progress


huitseeker (commented):
I think the commitment-related[^1] parts of :
- `kzg_commitment.rs`
- `pedersen.rs`
should stay out of the PCS module, because the traits they are dedicated to implementing are unrelated to the `EvaluationEngine`. They could be in their own `commitment` module if you like.

[^1]: I do not consider the implementation of `CommitmentKeyExtTrait` for the pedersen commitment to be commitment-related, and in fact it's only used by the IPA. So *that* trait implementation could move to the PCS module.

adr1anh (commented):
commenting to remove this PR from my "Review requested" list in GitHub

PR #374, opened 2024-03-21
chore: Move to the Ingonyama MSM

winston-h-zhang:
This PR switches the MSM on the bn256 curve to the Ingonyama MSM. We don't implement it directly, but rather use a wrapper repo https://github.com/lurk-lab/ingonyama-grumpkin-msm, which is similar to `grumpkin-msm`.


winston-h-zhang:
There is a little kink with the Ingonyama licenses -- should we make an upstream PR?

huitseeker:
> There is a little kink with the Ingonyama licenses -- should we make an upstream PR?

We should just open an issue to ingonyama, and fix it on our side with a `[[licenses.clarify]]` section in the deny.toml:
https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html

src/provider/bn256_grumpkin.rs
```diff
@@ -8,7 +8,7 @@ use digest::{ExtendableOutput, Update};
 use ff::{FromUniformBytes, PrimeField};
 use group::{cofactor::CofactorCurveAffine, Curve, Group as AnotherGroup};
 #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
-use grumpkin_msm::{bn256 as bn256_msm, grumpkin as grumpkin_msm};
+use grumpkin_msm::grumpkin as grumpkin_msm;
```
huitseeker, line 11: Why is it that the Grumpkin msm should still be on sppark? Doesn't ingonyama support grumpkin as well?

huitseeker (changes requested):
This is merged and we should use it:
https://github.com/ingonyama-zk/icicle/pull/379

src/provider/bn256_grumpkin.rs
```diff
@@ -48,6 +46,19 @@ fn bn256_msm(points: &[bn256::Affine], scalars: &[bn256::Scalar]) -> bn256::Poin
   }
 }
 
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+fn grumpkin_msm(points: &[grumpkin::Affine], scalars: &[grumpkin::Scalar]) -> grumpkin::Point {
+  cfg_if::cfg_if! {
+    if #[cfg(feature = "cuda")] {
+      let stream = ingonyama_grumpkin_msm::Config::new();
+      let cfg = ingonyama_grumpkin_msm::default_config(&stream.stream);
```
huitseeker, line 54: Could we make this a `static` in this configuration ?

src/provider/bn256_grumpkin.rs
```diff
@@ -35,6 +33,32 @@ pub mod grumpkin {
   };
 }
 
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+fn bn256_msm(points: &[bn256::Affine], scalars: &[bn256::Scalar]) -> bn256::Point {
+  cfg_if::cfg_if! {
+    if #[cfg(feature = "cuda")] {
+      let stream = ingonyama_grumpkin_msm::Config::new();
+      let cfg = ingonyama_grumpkin_msm::default_config(&stream.stream);
```
huitseeker, line 41: Could we make this a `static` to not create it on every call?

Cargo.toml
```diff
@@ -51,6 +51,7 @@ rayon-scan = "0.1.0"
 # grumpkin-msm has been patched to support MSMs for the pasta curve cycle
 # see: https://github.com/lurk-lab/grumpkin-msm/pull/3
 grumpkin-msm = { git = "https://github.com/lurk-lab/grumpkin-msm", branch = "dev" }
+ingonyama-grumpkin-msm = { git = "https://github.com/lurk-lab/ingonyama-grumpkin-msm", branch = "add-grumpkin-curve", optional = true }
```
huitseeker, line 54: Could we PR this small commit?

src/provider/bn256_grumpkin.rs
winston-h-zhang: I tried this, but ran into issues like
```
error[E0277]: `*mut icicle_cuda_runtime::bindings::CUstream_st` cannot be sent between threads safely
   --> src/provider/bn256_grumpkin.rs:39:16
    |
39  | static STREAM: Lazy<ingonyama_grumpkin_msm::Config> = Lazy::new(|| ingonyama_grumpkin_msm::Config::new());
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `*mut icicle_cuda_runtime::bindings::CUstream_st` cannot be sent between threads safely
    |
    = help: within `ingonyama_grumpkin_msm::Config`, the trait `Send` is not implemented for `*mut icicle_cuda_runtime::bindings::CUstream_st`
```
Dmytro [mentioned](https://folding-implementers.zulipchat.com/#narrow/dm/639273,651996,670525,693329-group/near/426492184) that the first stream creation will have some delay, but each time after that, the Icicle/CUDA runtime will be able to reuse them with little overhead. 
> Yeah, unfortunately the first CUDA call in any application always incurs some latency, it's one-time and fairly constant, 70 ms. in your case. If you try to do the same operations (create stream, alloc memory) again, the time will no longer be noticeable.

PR #380, opened 2024-04-23
Return ZK

tyshkor:


huitseeker (commented):
@tyshkor thanks for the PR that goes in an interesting direction. Unfrotunately, I don't manage to get it to compile. Here's a slightly modified branch I used to get past the trivial errors, such as unused imports:
https://github.com/huitseeker/arecibo/tree/return-zk
.. and that seems to not compile.

tyshkor:
> @tyshkor thanks for the PR that goes in an interesting direction. Unfrotunately, I don't manage to get it to compile. Here's a slightly modified branch I used to get past the trivial errors, such as unused imports: https://github.com/huitseeker/arecibo/tree/return-zk .. and that seems to not compile.

What OS are you using for it?

Divide-By-0:
hey! we noticed the last PR on this code was June 18, but it would be very useful for us to see the code upstreamed and maintained, so that we can have zk-based proving in systems like supernova/hypernova. do you have any timeline or sense of the priority of this so that can plan around that? would it be useful for us to help at all (i.e. fixing the tests and linting)?

PR #386, opened 2024-05-22
re-Adding lookups

wyattbenno777:
1. I moved @hero78119 's work into a new branch and rebased for the most recent version of dev branch.
2. Rather than having V2 I moved things into its own file lookup_ppsnark.rs.
3. Ran clippy.

Longer discussion can be found here:
https://github.com/lurk-lab/arecibo/pull/48

### DIFF
--- a/src/gadgets/lookup.rs
+++ b/src/gadgets/lookup.rs
@@ -1,0 +1,1023 @@
+//! This module implements lookup gadget for applications built with Nova.
+use std::cmp::max;
+use std::collections::btree_map::Iter;
+use std::collections::btree_map::Values;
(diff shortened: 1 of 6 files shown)


hero78119:
> rebased and PRed here #386

Hi @wyattbenno777 thanks for continuing this work and glad this can help somewhere ❤️ 

For memory consistency check protocol, https://eprint.iacr.org/2024/325.pdf propose more efficient protocol which eliminate the work of commit the `whole table` in the final compress snark. Please refer Appendix A for full protocols.

I believe once introduce this enhancement in compression snark then it will make this solution more robust.

